PR #2, opened 2019-04-12
Auto-unwrap args before calling function

sfentress:


sfentress:
This works on the expectation that these functions are always called with one argument, which is the case with all the functions so far, but I don't know blockly well enough to know if that's always the case.

PR #3, opened 2019-04-12
Adds a rotation transform before calculating tephra depth at gridX, Y

knowuh:
@sfentress 

This work isn't complete yet, but you can at least adjust the wind-direction in the controls and change the distrobution.

[#165290447]  Students can use a "set wind direction" block in their programs.

https://www.pivotaltracker.com/story/show/165290447

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14876,6 +14876,11 @@
         "punycode": "^2.1.0"
       }
     },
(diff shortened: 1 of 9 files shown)


sfentress:
Added the wind direction block functionality.

PR #8, opened 2019-04-22
Add settings panel

sfentress:
This adds a settings panel from [Dat.GUI](https://github.com/claus/react-dat-gui). At the moment this simply sets whether or not the erupt and paint blocks are required, or whether their actions are auto-run (as before), but eventually this could contain a number of authoring settings, and serialize itself to url parameters.

This is on top of #7 and can be rebased when that is merged.


src/components/app.tsx
```diff
@@ -120,8 +140,22 @@ export class AppComponent extends BaseComponent<IProps, IState> {
...
+          }
+        </DatGui>
+
       </App>
     );
   }
 
+  private toggleShowOptions = () => this.setState({showOptionsDialog: !this.state.showOptionsDialog});
+
+  private handleUpdate = (simulationOptions: any) => this.setState({ simulationOptions });
```
knowuh: Seems like eventually we could use `interface SimulationAuthoringOptions` for the  type  here.

sfentress: True! I'll put that in whatever PR comes next, so as not to need an extra rebase on the Maps work.

PR #16, opened 2019-05-09
Set VEI

sfentress:
Unfortunately there is no way of enforcing a specific range for `Number` inputs, if we are allowing variable inputs. (We could enforce a range with a `FieldNumber`, but then the user couldn't loop through values.)

Therefore this just notes the range on the block itself, and then clips in the code.

See e.g. Blockly's own color function, which accepts r, g, b values as variables and then does its own clipping under the hood: https://github.com/google/blockly/blob/d12374b3fbbc2b45579d738c2360856aa74000e6/generators/javascript/colour.js#L50


sfentress:
That's right, I forgot it was still an open question, and quickly added the check. We still need the check to ensure values >=0, is it would otherwise work fine with values <0 but make no sense to the students.

It's a one-line change if we chose to have no upper-limit.

PR #17, opened 2019-05-10
Atomic steps through blocks, and pausing on blocks

sfentress:
This makes it so that one click of the step button executes exactly one block. The step is executed at the same time that the block is highlighted, and then the execution is paused at the end of the block. This way the currently-highlighted block is the one being executed.

This also adds the ability for the simulation to pause and unpause the running of the blocks, to allow a block to take a custom amount of time when the user hits `Run`. Pausing and unpausing does nothing when the student is stepping, or if the function is executed outside of a run (e.g. on the Controls page).


sfentress:
:head_desk:

I just realized that the custom stepping code doesn't work for built-in blocks. Our main example is made up of all custom blocks, so I haven't thought about it.

I may have to switch to the STATEMENT_PREFIX version for now, and not worry about the highlighted block being the *next* step to execute. The Blockly devs have expressed interest in adding STATEMENT_POSTFIX, so hopefully we'd have another solution down the road.

Please pause this PR until I fix (probably on Monday).

sfentress:
@knowuh This is ready for PR now.

PR #20, opened 2019-05-21
Add data plot

sfentress:
This adds the data-plotting code. It's similar to what had been demoed previously, except

1. The block has been made more explicit, with x and y
2. The axes scale properly and the points sort, so we don't get oddities
3. The data is cleared if the user plots a different measurement, or if Reset is pressed

This is based off #19


sfentress:
Also a new author property has been added. To demo at https://geocode-app.concord.org/branch/add-data-plot-2/index.html, open the authoring and check off `showChart?`.

sfentress:
This is now rebased and building

PR #21, opened 2019-05-21
Improve controls

sfentress:
This adds custom tick marks and a string representation of the value to the sliders.

I can't tell if it's causing performance issues. I don't believe it ought to -- the library is very lightweight, as all the view code has to be added by hand -- but maybe the simulation painting is slowing down the rendering of the slider component, whereas before it was native?

If it seems to be a problem, we could also probably use a hybrid solution of a native range slider and custom ticks.

This is based on #20

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12910,6 +12910,32 @@
         "tinycolor2": "^1.4.1"
       }
     },
+    "react-compound-slider": {
+      "version": "2.0.0",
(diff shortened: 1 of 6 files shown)


knowuh:
@sfentress when you get a chance can you  rebase this on current master -- I have all your other PRs merged in there.  Thanks!

sfentress:
This is now rebased.

PR #22, opened 2019-05-30
sidebar visual indicators

Blarfnip:
This feature is being tracked in https://www.pivotaltracker.com/story/show/166159138
This can be demoed here:  http://geocode-app.concord.org/branch/sidebar-visual-indicators/index.html

Adds sidebar with visual indicators for column height, wind speed/direction, particle size, and miscellaneous data.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -213,6 +213,16 @@ export class AppComponent extends BaseComponent<IProps, IState> {
             map={ mapPath }
             isErupting={isErupting}
           />
+          <MapSidebarComponent
+            width={ 500 }
+            height={ 100 }
+            windSpeed={ windSpeed }
+            windDirection={ windDirection }
+            colHeight={ colHeight }
+            vei={ vei }
+            mass={ mass }
+            particleSize={ particleSize }
+          />
           { showCrossSection &&
             <CrossSectionComponent
               data={ data }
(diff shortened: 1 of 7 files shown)


src/components/app.tsx
```diff
@@ -212,6 +216,18 @@ export class AppComponent extends BaseComponent<IProps, IState> {
             map={ mapPath }
             isErupting={isErupting}
           />
+          { showSidebar && 
+            <MapSidebarComponent
+              width={ 500 }
```
lbondaryk: are we concerned about the use of magic hardcoded numbers in this way?

Blarfnip: when it comes to screen position data for components I don't necessarily agree with keeping them hardcoded, but I kept with the standard that was already set in the other components.

knowuh: Yeah, its true we have some hard-coded container widths at the moment.   This PR #18 includes work to dynamically resize content areas. We should get that PR merged soon. 

For now, this 500 will do in this branch, because the map has been explicitly set to that size.  We will revisit after #18 lands.

knowuh: This is magic number  comment is a good point.  Here we are setting width  and height  on the component itself.  In #18 we start to do more  dynamic layout, so I  think this is fine here for now.

PR #23, opened 2019-06-06
full screen mode

Blarfnip:
This feature is being tracked in: https://www.pivotaltracker.com/story/show/166388328
And can be demoed here: http://geocode-app.concord.org/branch/full-screen-mode/index.html

Adds a fullscreen button that enables native browser fullscreen for the entire activity.


Blarfnip:
@knowuh I just removed it from the store and combined it with the call from the app. I think it was in the store from when I was still experimenting.

PR #24, opened 2019-06-11
blockly resetting bug

Blarfnip:
So this works, but in order to make it work, I did a handful of probably not the best practice things in order to get around asynchronous file loading. If you think there is a better way to implement this, I'm all ears.

Demo here: http://geocode-app.concord.org/branch/blockly-resetting-bug/index.html


Blarfnip:
Fixed this in a better way in https://github.com/concord-consortium/geocode/pull/27

PR #27, opened 2019-06-12
fixed blockly resetting bug

Blarfnip:
Much simpler fix. Fixes issues with blockly code being non-persistent.

Should be demo-able here: http://geocode-app.concord.org/branch/blockly-resetting-simpler/index.html
(if travis isn't mad)

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -229,7 +229,7 @@ export class AppComponent extends BaseComponent<IProps, IState> {
                     { showControls && <Tab>Controls</Tab>}
                   </TabList>
                   { showBlocks &&
-                    <FixWidthTabPanel width={`${tabWidth}px`}>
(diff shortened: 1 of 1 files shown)


src/components/app.tsx
```diff
@@ -243,7 +243,7 @@ export class AppComponent extends BaseComponent<IProps, IState> {
                     <FixWidthTabPanel width={`${tabWidth}px`}>
                       <Code>
                         <SyntaxHighlighter>
-                          {js_beautify(code)}
+                          {js_beautify(code.replace(/endStep\(\)\;\n/g, "").replace(/startStep\(\'.*\'\)\;\n/g, ""))}
```
knowuh: ⬆  We also sneak in this fix in this  PR  👍  

Lets add a cleanup Chore in Pivotal Tracker to extract these regexs into a more general function.

PR #34, opened 2019-07-11
Leaflet experiments

Blarfnip:
This branch mainly consists of the leaflet conversion changes found in the following stories:
https://www.pivotaltracker.com/story/show/165250177
https://www.pivotaltracker.com/story/show/166936030
(kind of) https://www.pivotaltracker.com/story/show/166935883
https://www.pivotaltracker.com/story/show/166936092
https://www.pivotaltracker.com/story/show/166936186
https://www.pivotaltracker.com/story/show/166935672

On top of that, this conversion has also fixed a handful of untracked bugs

Demo here: http://geocode-app.concord.org/branch/leaflet-experiments/index.html


src/components/map-tephra-thickness-layer.tsx
```diff
@@ -0,0 +1,130 @@
...
+    particleSize: number;
+}
+
+interface IState {}
+
+@inject("stores")
+@observer
+export class MapTephraThicknessLayer extends BaseComponent<IProps, IState> {
+
+    // Not ideal, but geoJson does not update with state changes
```
knowuh: Interesting. This shouldn't be necessary.  Figuring out why a component isn't updating based on its properties can be tricky.

Each `<GeoJSON>` component in the collection  needs to have a unique `key` property, but these keys are usually simple indexes (something like  1- 8 in this example I think).

Any way, this isn't a big deal, and we can return to it later when we are polishing up.

src/stores/simulation-store.ts
```diff
@@ -245,13 +274,22 @@ export const SimulationStore = types
...
+      const vLng = self.volcanoLng;
+
+      self.windSpeed = self.stagingWindSpeed;
+      self.windDirection = self.stagingWindDirection;
+      self.colHeight = self.stagingColHeight;
+      self.mass = self.stagingMass;
+      self.vei = self.stagingVei;
+      self.particleSize = self.stagingParticleSize;
+
+      // This eruption data is no longer calculated correctly or used
```
knowuh: Do you think we are ever going to want to use this data again?  

If its only used for the old grid squares then maybe we can get rid of it.

package.json
```diff
@@ -113,11 +113,18 @@
   },
   "dependencies": {
     "@inlet/react-pixi": "^0.5.7",
+    "@types/d3": "^5.7.2",
+    "@types/d3-contour": "^1.3.0",
+    "@types/react-leaflet": "^2.2.1",
     "@types/recharts": "^1.1.16",
     "color": "^3.1.0",
+    "d3": "^5.9.7",
```
knowuh: As you mentioned, it seems odd to include all of D3 in order to do the contour work.  

OTOH, it looks great. 
Also OTOH: Now that we are including D3 maybe we can lean on it in other views.

Are you interested in trying to implement the contours without D3?

src/components/map-component.tsx
```diff
@@ -60,66 +110,120 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
-            width={60}
-            height={60} />
-          <VolcanoEmitter
-            config={AshConfig.config as unknown as EmitterConfig}
-            imagePath={AshConfig.image}
-            x={volcanoPos.x + 20}
-            y={volcanoPos.y + 20}
-            windDirection={windDirection}
+          <TileLayer
+              url="https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png"
```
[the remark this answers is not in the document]
dstrawberrygirl: We have a tile server to cache earthquake data (since USGIS database occasionally pops offline during gov. shutdowns) but we do not use a tile server for the map tiles themselves. The OSM tiles for street maps we use in Seismic Explorer also include an optional parameter for scale that comes from window.devicePixelRatio - on Retina displays, map tiles will appear fuzzy unless you add that parameter (@2x) works well if I recall correctly. 

Example usage here: https://github.com/concord-consortium/seismic-explorer/blob/master/js/map-layer-tiles.js. 

Result is seen https://maps.wikimedia.org/osm-intl/8/66/118.png compared with https://maps.wikimedia.org/osm-intl/8/66/118@2x.png - when rendered on retina, that higher resolution image looks super sharp. Let me know if you need a hand here!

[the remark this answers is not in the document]
dstrawberrygirl: As an addendum to that - you can use the 2x images on all devices without negative rendering effects - the bandwidth used will be a little higher but it's not a huge map so probably not a big issue. Actual devicePixelRatios can vary, so you could switch to using 2x on any ratio greater than 1

src/stores/simulation-store.ts
Blarfnip: I removed a lot of the unused variables and put console warning messages in for the function calls until we get to changing the blockly code

PR #35, opened 2019-07-18
167291916 ruler tool

Blarfnip:
Story: https://www.pivotaltracker.com/story/show/167291916
(https://www.pivotaltracker.com/story/show/166936300)

Added ruler tool and changed a bunch of styling on various buttons and icons to make things look a little better.

Demo: http://geocode-app.concord.org/branch/ruler-tool/index.html

### DIFF
--- a/src/components/cross-section-component.tsx
+++ b/src/components/cross-section-component.tsx
@@ -22,6 +22,8 @@ interface IState {
   isSelecting: boolean;
 }
 interface IProps extends IBaseProps {
+  showCrossSectionSelector: boolean;
+  isSelectingCrossSection: boolean;
   height: number;
   width: number;
   volcanoLat: number;
@@ -65,6 +67,14 @@ export class CrossSectionComponent extends BaseComponent<IProps, IState>{
 
   public componentDidUpdate(prevProps: IProps) {
     this.recomputeMetrics();
+
+    if(this.props.isSelectingCrossSection !== prevProps.isSelectingCrossSection) {
+      this.stores.setCrossSectionSelectorVisibility(this.props.isSelectingCrossSection);
+    }
+
+    if (!this.stores.showCrossSectionSelector && this.state.isSelecting) {
+      this.setState({isSelecting: false});
+    }
   }
 
   public render() {
@@ -83,17 +93,16 @@ export class CrossSectionComponent extends BaseComponent<IProps, IState>{
       windDirection,
       colHeight,
       mass,
-      particleSize
+      particleSize,
+      isSelectingCrossSection
     } = this.props;
     const { width } = this.metrics;
 
     return (
       <CanvDiv ref={this.ref}>
         {hasErupted && <ContainerDiv>
-          {!isSelecting && <Button onClick={this.selectButton}>Draw a cross section line</Button> }
-          {isSelecting &&
+          {isSelectingCrossSection &&
           <ContainerDiv>
-            <Button onClick={this.cancel}>Cancel</Button>
             <Stage
               width={width}
               height={height}
--- a/src/components/cross-section-component.tsx
+++ b/src/components/cross-section-component.tsx
@@ -18,9 +18,8 @@ const ContainerDiv = styled.div`
   height: flex;
 `;
 
-interface IState {
-  isSelecting: boolean;
-}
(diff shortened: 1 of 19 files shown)


src/components/cross-section-component.tsx
```diff
@@ -65,6 +67,14 @@ export class CrossSectionComponent extends BaseComponent<IProps, IState>{
 
   public componentDidUpdate(prevProps: IProps) {
     this.recomputeMetrics();
+
+    if (this.props.isSelectingCrossSection !== prevProps.isSelectingCrossSection) {
+      this.stores.setCrossSectionSelectorVisibility(this.props.isSelectingCrossSection);
```
knowuh: Dispatching an action in `componentDidUpdate` is valid, but if there is another way to do it, I would.

`props.isSelectingCrossSection` is passed as a prop from `app.tx` which is reading that value from the store.  

Is there a way to handle these cases in the stores actions or possibly a computed property?

src/components/cross-section-component.tsx
```diff
@@ -65,6 +67,14 @@ export class CrossSectionComponent extends BaseComponent<IProps, IState>{
...
+    }
+
+    if (!this.stores.showCrossSectionSelector && this.state.isSelecting) {
+      this.setState({isSelecting: false});
```
knowuh: We are managing `isSelecting` in local state, how does this overlap with with `props.isSelectingCrossSection`.   

I feel like we can simplify things. I would try to manage these items in one place. Probably the application state. (Stores)

src/components/map-component.tsx
```diff
@@ -96,6 +101,13 @@ export class MapComponent extends BaseComponent<IProps, IState>{
     this.forceUpdate();
   }
 
+  public componentDidUpdate(prevProps: IProps) {
```
knowuh: My comments about `CrossSectionComponent` state management my apply to this component too, seems like a similar pattern.

src/components/map-component.tsx
```diff
@@ -169,44 +190,73 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
+          onCrossSectionClick={this.onCrossSectionClick}
+        />
       </CanvDiv>
     );
   }
 
+  private onRulerClick = () => {
+    this.setState({showRuler: !this.state.showRuler});
+    this.stores.setCrossSectionSelectorVisibility(false);
+    this.stores.setIsSelectingCrossSection(false);
```
knowuh: `onRulerClick` changes the local state, and then invokes two actions on the stores.  

Because the cross-section view and ruler views need to know about it each other, it may be easier to do this with explicit actions & props in the stores.

src/components/ruler-draw-layer.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const { volcanoLat, volcanoLng } = this.stores;
+    const point1 = L.latLng(volcanoLat, volcanoLng);
+    const point2 = L.latLng(pointLat, pointLng);
+    const point3 = L.latLng(volcanoLat, pointLng);
+    const p1Icon = getCachedCircleIcon("");
+    const p2Icon = getCachedCircleIcon("");
+
+    const xDist = getDistanceFromLatLonInKm(point1, point3);
+    const yDist = getDistanceFromLatLonInKm(point2, point3);
+    const adjustedXDist = point3.lng < point1.lng ? -1 * xDist : xDist;
```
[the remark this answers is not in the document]
Blarfnip: The ? operator is actually being used to do the opposite. The getDistance function returns an unsigned distance, so I have to put the signs back on using that statement.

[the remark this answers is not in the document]
knowuh: Oops, sorry. I see now.

src/components/map-component.tsx
```diff
@@ -186,7 +170,7 @@ export class MapComponent extends BaseComponent<IProps, IState>{
           zoomControl={true}
           doubleClickZoom={true}
           scrollWheelZoom={true}
-          dragging={!showCrossSectionSelector}
+          dragging={!hasErupted}
```
knowuh: This is the only line I find strange.  

Do we really want to prevent dragging if the volcano has erupted?

PR #38, opened 2019-07-29
separated erupt and paint into two blocks

Blarfnip:
I'm not the happiest with this solution, because it seems like the stores are getting cluttered up by multiple copies of the simulation parameters. 

Demo here: http://geocode-app.concord.org/branch/color-map-block/index.html


Blarfnip:
I changed the default code, so the Travis automatic tests are broken. All the code compiles and runs fine from my testing.

PR #42, opened 2019-10-10
Clean up

sfentress:
This is a collection of small fixes, and a small bugfix that makes the nested loops example work.

This doesn't make any structural changes to the work that was done over the summer.

The one thing I have a question about is why the simulation of the eruption (the tephra calculation) was changed. Previously the `simulation-store` ran the calculation once, on erupt, and calculated all the tephra values, and then the components simply looked at those values. Now the various views themselves actually run `gridTephraCalc` on `render`. I can't quite work out the purpose of this, unless maybe there was some difficulty with getting MobX Statetree to properly rerender views when nested data structures were being updated? Just a guess.

(You can see where the calculation was added to `map-tephra-thickness-layer` [here](https://github.com/concord-consortium/geocode/commit/355db7a8dca35d9f102ef02317171519460b6237#diff-ab4a64dbe6d8071e654962f91e21f1a5), and it gets added to other views and removed from the store elsewhere.)

However, it all seems to work, so unless we see inefficiencies with calculating multiple times, I'm ok leaving it.


sfentress:
Forgot to commit unused PIXI components.

We could bring these back later if we ever want to use them again, but at the moment these are all orphans.

sfentress:
I'm going to bring this in. @knowuh if you think you have any thoughts about the simulation calculation question above, let me know, otherwise I'll just assume it relates to MobX difficulties or something.

mklewandowski:
Thanks for bringing this in @sfentress.  I'll have the first UI PR in later today.  I didn't want to submit it until I could rebase onto these changes.

PR #44, opened 2019-10-23
167253922 left panel tabs ui

mklewandowski:
Update the styling of the left panel tabs to match UI specs.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -72,6 +72,7 @@ const App = styled.div`
 const Row = styled.div`
   display: flex;
   width: 100%;
+  height: 100%;
   justify-content: space-between;
   align-items: center;
   flex-direction: row;
@@ -92,7 +93,10 @@ interface ISim {
 const Code = styled.div`
   max-height: 400px;
   overflow: auto;
-  padding: 1em;
+`;
+
+const Syntax = styled(SyntaxHighlighter)`
+  margin: 0px;
 `;
 
 const FullscreenButton = styled(StyledButton)`
@@ -311,7 +315,11 @@ export class AppComponent extends BaseComponent<IProps, IState> {
               { showControls && <Tab>Controls</Tab>}
             </TabList>
             { showBlocks &&
-              <FixWidthTabPanel width={`${tabWidth}px`} forceRender={true}>
+              <FixWidthTabPanel
+                width={`${tabWidth}px`}
+                forceRender={true}
+                tabcolor={"#DDEDFF"}
+              >
                 <BlocklyContainer
                   width={blocklyWidth}
                   height={blocklyHeight}
@@ -329,16 +337,22 @@ export class AppComponent extends BaseComponent<IProps, IState> {
               </FixWidthTabPanel>
             }
             { showCode &&
-              <FixWidthTabPanel width={`${tabWidth}px`}>
+              <FixWidthTabPanel
+                width={`${tabWidth}px`}
+                tabcolor="#BBD9FF"
+              >
                 <Code>
-                  <SyntaxHighlighter>
+                  <Syntax>
                     {js_beautify(code.replace(/endStep\(\)\;\n/g, "").replace(/startStep\(\'.*\'\)\;\n/g, ""))}
-                  </SyntaxHighlighter>
+                  </Syntax>
                 </Code>
               </FixWidthTabPanel>
             }
             { showControls &&
-              <FixWidthTabPanel width={`${tabWidth}px`}>
+              <FixWidthTabPanel
+                width={`${tabWidth}px`}
+                tabcolor="#FFCA79"
+              >
                 <Controls
                   showWindSpeed={showWindSpeed}
                   showWindDirection={showWindDirection}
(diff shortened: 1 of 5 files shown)


src/components/tabs.tsx
```diff
@@ -6,41 +6,97 @@ import {
...
+}
+type TabInfo = {
+  [tab in SectionTypes]: {
+    name: string;
+    index: number;
+    backgroundColor: string;
+    hoverBackgroundColor: string;
+  }
+};
+const kTabInfo: TabInfo = {
```
mklewandowski: I'm open to naming convention ideas here if people don't like `kTabInfo` (this is convention in CLUE and some of our other apps).

src/components/controls.tsx
```diff
@@ -11,6 +11,8 @@ const StyledControls = styled.div`
   flex-direction: column;
   justify-content: flex-start;
   align-items: flex-start;
+  background-color: #F0F0F0;
```
mklewandowski: We probably need a central location for commonly used colors from the UI spec.  I'm open to ideas.

src/components/app.tsx
```diff
@@ -298,47 +311,103 @@ export class AppComponent extends BaseComponent<IProps, IState> {
...
+                  backgroundcolor={this.getTabColor(SectionTypes.CODE)}
+                  backgroundhovercolor={this.getTabHoverColor(SectionTypes.CODE)}
+                >
+                  {this.getTabName(SectionTypes.CODE)}
+                </Tab>
+              }
+              { showControls &&
+                <Tab
+                  borderwidth={this.state.tabIndex !== kTabInfo.controls.index ? "2px" : "0px"}
+                  leftradius={this.state.tabIndex === (kTabInfo.controls.index - 1) ? "10px" : "0px"}
```
mklewandowski: Note that I went with passing the value of the CSS property in via the styled component props.  Another pass is to pass in a value that is used in some computation or comparison to obtain the property value inside of the component.  I like the flexibility of passing in an arbitrary property value without having to create a complex calculation inside of the styled component CSS.

src/components/controls.tsx
knowuh: Seems like a good idea to have a file specific for those style-related vars at a minimum.

We could also look into pulling those vars in via `<ThemeProvider>` but that may be overkill for what we are doing here.

src/components/app.tsx
knowuh: Generally a fan of semantic properties, so having a property like `isTerminal` or  `isFinal` `isLast` and then having the component do the right thing (based on its own metrics) would be my preference with this sort of thing, but I understand your concerns too.

src/components/controls.tsx
mklewandowski: I'll look at pulling all of the colors out of the spec and putting them in one module in a separate PR.  But I think we both agree that it would be useful.

src/components/app.tsx
mklewandowski: The semantic route was what I initially went with and what I would prefer for a typical React component (let the component decide what to do based on the props), but that model seems a little less intuitive with styled components and wasn't in line with some of the examples in the documentation.  This is especially true when passing in things like colors or dealing with CSS properties that have more than a binary state (the CSS code in the styled component gets ugly) like we do with some of these complex border stylings needed to get the tabs to match the spec.  However, I'm game to tweak this if others think semantic properties are the right direction.

sfentress:
I'll let you guys keep conversing, as you have more intelligent things to say about styled components than I do, but this looks generally good to me.

src/components/app.tsx
knowuh: I wonder if you could just wrap the styled component with a simple functional component wrapper that does those calculations, and export that?  

I wouldn't sweat this though, its not going to stall a PR :)

src/components/app.tsx
```diff
@@ -298,47 +311,103 @@ export class AppComponent extends BaseComponent<IProps, IState> {
...
+    kTabInfo.code.index = showCode ? kTabInfo.blocks.index + 1 : -1;
+    kTabInfo.controls.index = showControls ? (showCode ? kTabInfo.code.index + 1 : kTabInfo.blocks.index + 1) : -1;
+    let currentTabType = SectionTypes.BLOCKS;
+    if (showCode && this.state.tabIndex === kTabInfo.code.index) {
+      currentTabType = SectionTypes.CODE;
+    }
+    if (showControls && this.state.tabIndex === kTabInfo.controls.index) {
+      currentTabType = SectionTypes.CONTROLS;
+    }
+
```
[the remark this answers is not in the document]
mklewandowski: Yeah, that works as long as we put some extra protection around `currentTabType` being `undefined`.  `currentTabType` always held a valid value before, but now can be `undefined` as a user turns tabs on and off from the menu.

src/components/app.tsx
mklewandowski: I tweaked the props to use semantic names and let the component determine its own styling.  I was a little thrown off by having to pass in strings (and non-camel case), but maybe someone with more experience with styled components in Typescript knows a better pattern.  Nonetheless, I think it's probably ok for now.

PR #45, opened 2019-10-24
Save code to LARA

sfentress:
This can be demo'd at http://authoring.concord.org/activities/10028/pages/127921/

As part of this (as it kind of came along for free) I also added Save and Load from local state to the authoring popup menu, since this is something we had wanted before. This is probably more of a gimmick than a useful feature right now, but there is no reason we couldn't expand it to save to multiple file names, and/or go for the download/upload file route, if a non-LARA authoring solution is still desired.

The authoring buttons can be tested here: http://geocode-app.concord.org/branch/save-code-to-lara/index.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7418,6 +7418,11 @@
       "integrity": "sha1-xg7taebY/bazEEofy8ocGS3FtQE=",
       "dev": true
     },
+    "iframe-phone": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/iframe-phone/-/iframe-phone-1.2.0.tgz",
+      "integrity": "sha1-6RP79PeWTu7My3u5rLqtVWIeHGQ="
+    },
     "ignore": {
       "version": "3.3.10",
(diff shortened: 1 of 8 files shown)


sfentress:
Ah, interesting catch. I hadn't seen that, but it makes sense. I had noticed with the local storage save/load that I needed to introduce a 100ms pause between loading two different things, or Blockly would put up two containers.

LARA must be first loading the app, and then sending the learner data within that timeframe.

I can throttle LARA, or see if I can figure out a real solution to the double-loading issue.

sfentress:
The double-container bug should be fixed now.

knowuh:
@sfentress:  This looks good, but it doesn't seem like I am getting my blocks loaded when I return tp the LARA page.

sfentress:
Ok.... Turned out to be an annoying race condition, where authored code would start to be requested from an external file, the LARA saved data would come in, and then the authored code would finish its request. Depended on timing, so I didn't see it at first.

Note, when you test on LARA now, you may have several copies of the same code stacked on top of each other from previous double-loads. Clear that out if so, and then test saving and loading.

PR #46, opened 2019-10-29
Save and load authored blockly code

sfentress:
This is a small addition to save the authored blockly xml, using the available LARA features, and piggybacking off the student save-data model.

I was going to include the authoring support for the rest of the model parameters, but that is becoming a much bigger commit so I thought it wise to get the small one out first.


mklewandowski:
@sfentress what's the easiest way to test these changes?

PR #48, opened 2019-10-30
167254984 controls tab ui

mklewandowski:
This PR implements the controls panel styling as specified in the UI spec.  It also creates a path for SVG integration that can be used for future icon/art integration.  

A few caveats that are worth discussing before bringing this in (also posted on PT):
-There are a few icon colors that are missing (I'd rather get the colored icons in cases where there is nuanced icon colors rather than trying to create a path where we dig into the SVG and adjust path colors).  However, the icons themselves are slotted in and can be easily replaced once we have additional assets.
-Based on discussion in the spec, there will obviously be additions, modifications, or possibly deletions from the set of controls. For this PR I tried to address only the styling and have not made changes to the controls themselves (so the controls that were there are still there). I feel the best move is to wait until the control set is locked down and then put in the final set of controls in another branch and PR (this PR is already getting bulky and I don't want to slow down other work as we wait on updated specs).  If others would like to see those changes in this PR, however, I can continue working on them here.


mklewandowski:
Based on review and discussion on PT, we'll handle changes to the actual slider controls and functionality in other stories.

PR #57, opened 2019-11-04
Convert eruption mass to ejected volume

mklewandowski:
Change the eruption mass slider (in kg) to an ejected volume slider (km^3).  Keep underlying calculations in kg.  This only changes the user-facing slider in the control panel (and the options in the model options panel).  

Note: I did not change any of the blockly blocks that may incorporate the mass in kg - this might be something that we want to consider.


mklewandowski:
@sfentress The changes to the setMass block are defined in another story (PT id 169506779).  My hope was to complete this story and then finish the VEI slider story and the setMass block story.

PR #59, opened 2019-11-06
Add VEI slider

mklewandowski:
This PR adds a VEI slider to the controls panel that ranges from 1 - 8.  Slider can be shown/hidden from the model options.  When adjusted, slider also updates the ejected volume and the column height so they are in sync with the selected VEI value.  Full VEI widget with updated display based on VEI slider value will be handled in a subsequent PR.

### DIFF
--- a/src/components/controls.tsx
+++ b/src/components/controls.tsx
@@ -47,17 +47,21 @@ const ControlLabel = styled.label`
   margin-top: 6px;
 `;
 
+interface ValueContainerProps {
+  width?: number;
+}
 const ValueContainer = styled.div`
   display: flex;
   flex-direction: column;
   justify-content: center;
   align-items: center;
-  width: 104px;
+  width: ${(p: ValueContainerProps) => `${p.width ? `${p.width}px` : "104px"}`};
   height: 80px;
   margin-left: auto;
   padding: 2px;
   border-radius: 7px;
   background-color: #FFDBAC;
+  font-size: 12px;
 `;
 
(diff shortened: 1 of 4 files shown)


src/components/controls.tsx
```diff
@@ -355,6 +414,24 @@ export class Controls extends BaseComponent<IProps, IState> {
     this.stores.setParticleSize(size);
   }
 
+  private changeVEI = (vei: number) => {
+    this.stores.setVEI(vei);
+  }
+
+  private getVEIDescription = (vei: number) => {
+    switch (vei) {
```
kswenson: Would recommend a map rather than a switch statement.

mklewandowski: Good call.  I'm going to move this and the VEI column height into a standardized object that we can pull values from and add future properties to if needed.

mklewandowski:
@sfentress Good points.  I would vote to address some of that in a new story since there are other related issues tied to ensuring that controls have proper ranges (which I agree is complicated now that one control can adjust the value of other controls).  For example, the column height slider has a min value of 1, but if VEI is set to 1, it tries to set a corresponding column height value of .1 (which cannot be shown on the column height slider).  Similarly, the VEI block tells the user to enter a value from 0 - 8 instead of the 1 - 8 allowed on the slider.  I think, in general, we should look at these value ranges and add new stories for cases where a control doesn't have the proper range or should update it's current value based on some other app change.  I'll post some of this to Slack.

PR #60, opened 2019-11-06
169505530 vei widget

mklewandowski:
This PR adds a VEI slider widget which is used in the left control panel and in the right panel beneath the map.  Some notes:
 - we don't yet have an updated right panel with a space to slot in the widget so the right panel integration is temporary
 - I added `styled-containers.tsx` to place some simple containers that are reused across modules
 - `widget.ts` contains some basic shared information about the widgets and their styling.  We can add to this as we add more widgets.


mklewandowski:
@sfentress I'm going to bring this in for now, but as I work on the other widgets, I'll take a look at the SVG integration and circle back to this one if we can slim down the number of imported SVGs.

PR #65, opened 2019-11-08
Add column height widget

mklewandowski:
This PR adds the column height widget to the left controls panel and to the right sidebar beneath the map (note that the positioning here is temporary as we wait for the right panel styling updates to be completed).


sfentress:
Ignore my previous comment, I recall now the the left widget looks at the staging value, and the right widget looks at the last eruption value. :+1:

PR #83, opened 2019-11-13
Update button states and styling

mklewandowski:
Small fixes to the button states and styling on the map:
 - only show cross section button if we are on the cross section tab
 - style cross section button to match cross section tab
 - have ruler and cross section buttons have an "activated" (AKA "on") color state for when the user is in the process of using the button feature (measuring with ruler or drawing cross section line)


llzes:
The styling looks fantastic! Looks good to me ✨

PR #88, opened 2019-11-13
Fix widgets so that they respect min and max values

mklewandowski:
This PR updates the ejected volume, wind, VEI, and column height widgets so that they enforce min and max values for the visual, icon portion of the widget.  Using the blockly blocks, a user can specify eruption values that are outside of the expected range of values used within the widgets.  This PR allows the block specified value to be displayed in the text field within the widget, but also constrains the min and max values used when drawing the icon portion of the widget (prevents things like an ejected volume that is drawn outside the box or a wind speed arrow that is extremely long).  In the future we might want to enforce min and max values for the blockly block, but at least for now we can prevent the widgets from drawing bizarre or confusing representations of the widget values.

### DIFF
--- a/src/components/column-height-widget.tsx
+++ b/src/components/column-height-widget.tsx
@@ -52,12 +52,14 @@ export default class ColumnHeightWidget extends PureComponent<IProps, IState> {
   };
 
   public render() {
+    const minColumnHeight = .1;
     const maxColumnHeight = 25;
     const maxGuageHeight = 48;
     const maxCoverHeight = 50;
     const { columnHeightInKilometers, type } = this.props;
-    const guageHeight = maxGuageHeight * (columnHeightInKilometers) / maxColumnHeight;
-    const coverHeight = maxCoverHeight - (maxCoverHeight * columnHeightInKilometers / maxColumnHeight);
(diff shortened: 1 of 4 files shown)


src/components/column-height-widget.tsx
```diff
@@ -52,12 +52,14 @@ export default class ColumnHeightWidget extends PureComponent<IProps, IState> {
...
   public render() {
+    const minColumnHeight = .1;
     const maxColumnHeight = 25;
     const maxGuageHeight = 48;
     const maxCoverHeight = 50;
     const { columnHeightInKilometers, type } = this.props;
-    const guageHeight = maxGuageHeight * (columnHeightInKilometers) / maxColumnHeight;
-    const coverHeight = maxCoverHeight - (maxCoverHeight * columnHeightInKilometers / maxColumnHeight);
+    const constrainedColumnHeight = Math.min(Math.max(columnHeightInKilometers, minColumnHeight), maxColumnHeight);
+    const guageHeight = maxGuageHeight * (constrainedColumnHeight) / maxColumnHeight;
```
kswenson:
```suggestion
    const guageHeight = maxGuageHeight * constrainedColumnHeight / maxColumnHeight;
```

PR #89, opened 2019-11-13
Minor fixes to right tab display

mklewandowski:
Minor fixes to the right tabs to improve tab display so they more closely match the UI spec:
- Data index was incorrectly computed - caused incorrect tab borders to be displayed
- Tab back was incorrectly placed


src/components/app.tsx
```diff
@@ -212,7 +212,7 @@ export class AppComponent extends BaseComponent<IProps, IState> {
     kRightTabInfo.conditions.index = showConditions ? 0 : -1;
     kRightTabInfo.crossSection.index = showCrossSection ? kRightTabInfo.conditions.index + 1 : -1;
     kRightTabInfo.data.index = showData
-      ? (showCrossSection ? kRightTabInfo.conditions.index + 1 : kRightTabInfo.conditions.index + 1)
+      ? (showCrossSection ? kRightTabInfo.crossSection.index + 1 : kRightTabInfo.conditions.index + 1)
```
sfentress: Oops, good catch

PR #92, opened 2019-11-14
Update tephra calculations

sfentress:
This updates the calculations to be a rewrite of the Python program provided by Robert Constantinescu. This model integrates over many phi classes (particle sizes) and over many cells of a simulated large eruption disk.

The main update is a direct re-write of the code, and is confirmed using test that use expected values from the Python program's output.

The program was then changed slightly as follows:

1. Number of simulated phi classes is settable, and the default lowered to 7 classes
2. Size of eruption cells to sample from is settable, and the default size increased to 3000

These changes were to balance speed and accuracy, an a comparison can be seen at
https://docs.google.com/spreadsheets/d/12EEJpJAnqk5pMMoa4cdtDSqUv7uiRPIPAGEIev1SJfQ/edit?usp=sharing

An additional change was added

3. At eruptions of VEI < 4, a small eruption disk is modeled

This is from a comment from Robert over email that the mushroom cloud simulation is only applicable at VEI >= 4. The comparison between the two models can also be seen at the above link.

### DIFF
--- a/src/components/controls.tsx
+++ b/src/components/controls.tsx
@@ -173,11 +173,11 @@ export class Controls extends BaseComponent<IProps, IState> {
                 <ControlLabel>Column Height (km)</ControlLabel>
                 <HorizontalContainer>
                   <RangeControl
-                    min={.1}
+                    min={.5}
                     max={25}
                     value={stagingColHeight / 1000}
                     step={.1}
-                    tickArray={[.1, 5, 10, 15, 20, 25]}
+                    tickArray={[.5, 5, 10, 15, 20, 25]}
                     width={this.props.width - 220}
                     onChange={this.changeColumnHeight}
                   />
--- a/src/tephra2.ts
+++ b/src/tephra2.ts
@@ -2,6 +2,8 @@
 // Chuck Connor  & Noah Paessel 2/2019
 // using tephra2 algorthim (equation of Lim et al. (2008))
 import { rotateGridPoint } from "./utilities/grid-transform";
+import { Ipoint } from "./interfaces";
+
 const tephraCalc = (
   x: number,
   y: number,
@@ -80,7 +82,7 @@ const tephraCalc2 = (
     return masLoading;
 };
 
-const gridTephraCalc = (
+const gridTephraCalc2 = (
   gridX: number,
   gridY: number,
   coneGridX: number,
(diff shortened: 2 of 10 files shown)


src/tephra2.ts
```diff
@@ -2,122 +2,299 @@
...
+
+    const diameter = getDiameter(phi);
+
+    const vL = (g * diameter ** 2 * rho) / (18 * airVisc);
+    const vI = diameter * ((4 * g ** 2 * rho ** 2) / (225 * airVisc * rhoA)) ** (1 / 3);
+    const vT = ((3.1 * rho * g * diameter) / (rhoA)) ** 0.5;
+
+    const reL = (diameter * rhoA * vL) / airVisc;
+    // const reI = (diameter * rhoA * vI) / airVisc;
+    const reT = (diameter * rhoA * vT) / airVisc;
```
knowuh: I assume that this came from Robert, and I can't claim to be able to validate this by looking :)  

Some of the constant values like 8200, 18, 31, and 255 aren't obvious to me for example.

That being said, I am not too concerned with understanding this.  Having the tests is enough for our purposes here.

src/components/controls.tsx
```diff
@@ -173,11 +173,11 @@ export class Controls extends BaseComponent<IProps, IState> {
                 <ControlLabel>Column Height (km)</ControlLabel>
                 <HorizontalContainer>
                   <RangeControl
-                    min={.1}
+                    min={.5}
                     max={25}
                     value={stagingColHeight / 1000}
                     step={.1}
-                    tickArray={[.1, 5, 10, 15, 20, 25]}
+                    tickArray={[.5, 5, 10, 15, 20, 25]}
```
mklewandowski: There might be a couple of related changes if the new min val of the col height slider is .5.  

Could we also change the min val of the widget to .5 here?
https://github.com/concord-consortium/geocode/blob/master/src/components/column-height-widget.tsx#L55

Not sure if it also has an impact on the VEI related col height val.  I'm assuming no, but worth a look:
https://github.com/concord-consortium/geocode/blob/master/src/utilities/vei.ts#L9

sfentress: :+1:

sfentress:
@knowuh I'm going to go ahead and bring this in. If anyone needs to compare to the previous version, the last non-master branch that had the old code was `restore-force-render`, so the old tephra calc can be permanently seen at http://geocode-app.concord.org/branch/restore-force-render/index.html

PR #95, opened 2019-12-03
Np experiments

knowuh:


sfentress:
Hi @knowuh 

I've looked through these, and most of this looks great and ready to be brought in.

I think you mentioned that this was a bunch of merging things into one PR, so you could get things in front of students. As I went through this, I felt like these fell into a few different categories:

### These are all small changes, and can be trivially merged in:

* Change block colors 00e631d
* Add wind data. d46ca1d
* Add mockup assets 66cd8b6
* Changes to toolboxes. abdc0c4
* rename `setWinddirection` function 679129a

### These are larger features

* Add `run-simulation` code block. Recoller toolbox & blocks. 0b63069
* Configure lesson specific toolboxes 0a4fa3d

The code looks fine to me, and these seem like they've been used a bunch, at least by you. Do you think any more testing needed?

* Add optional markers to scenarios 5e37736

This looks great to me.

### Unsure of the status of these commits:

* WIP: Example Sampling blocks 6eab3e6
* WIP Experiment -- add some list & function blocks 38d37d4
* Experimental histogram component. Not used really. bfea973

Are these ready to be merged in?

### Possibly made unnecessary by #100

* Author can set VEI by manipulating sliders. 0d2b492

----

For all of these, I'd be happy to do some rebasing of this PR into smaller groups, if we find that some things need to be brought in and others not.

Pinging @mklewandowski and @kswenson since they wanted to be kept in the loop on this.

knowuh:
Sam, thanks for weeding through this mess. 

This branch worked well in our trials, so I am comfortable with it from that perspective.

As you point out, these commits could/should be removed, because they really were experiments:

* WIP: Example Sampling blocks 6eab3e6
* WIP Experiment -- add some list & function blocks 38d37d4
* Experimental histogram component. Not used really. bfea973

If #100 fixes the same issue as 0d2b492 (and does it in a better way) then that is fantastic, and we can skip 0d2b492 too.

Thanks again

sfentress:
Ok, I'm going to do some rebasing and will split this into three PRs for record-keeping, the trivial set, the Run Simulation changes and the Markers, and will approve and merge them in (unless anyone has any objections or comments).

sfentress:
I'm going to close this, as the three associated PRs are merged in.

This branch does container some potentially very useful WIP work on historic data and lists, so it's worth keeping it around and remembering that it exists here.

PR #98, opened 2020-01-09
Author can set VEI by manipulating sliders.

knowuh:
Hello @sfentress --

Here is the change that let the `stagingVEI` param (and others) be set using the control sliders in authoring mode.

I wanted you to look it, because we were discussing it this morning.  Also, I recognize that you went out of the way to ensure that only certain simulation parameters were saved on serialization.  This commit basically undoes all that work.

This is a change that I would just want to use while in the classroom this week.  Chris has tested it, and it works for his needs.

Let me know if you have any reservations about committing this.  And thanks for the review.

### DIFF
--- a/src/stores/stores.ts
+++ b/src/stores/stores.ts
@@ -107,9 +107,15 @@ export const deserializeState = (serializedState: SerializedState): IStoreish =>
 };
 
 export function updateStores(state: IStoreish) {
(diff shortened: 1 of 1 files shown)


sfentress:
I am going to close this, as we have merged in #100. If there are issues not addressed by that, we can bring this back in.

PR #99, opened 2020-02-07
Smooth tephra calculation

sfentress:
This is a first-pass at improving the smoothness of the tephra deposits at high wind speeds. It modifies the existing tephra calculation by

1. Adding new phi classes between phi 1 and phi 3
2. Iterating the calculation over three different column heights, and 80%, 90% and 100% of the total height

Eruption with 15 m/s wind prior to changes:

Screen Shot 2020-02-05 at 1 43 15 PM

Same eruption with new phi classes:

8-phi-1-height2

New phi classes and new height integration:

8-phi-3-height

Note:

The height integration is the most computationally-intensive part (it simply repeats the whole calculation for every height) and has less of a visible effect than the new phi classes. It is, however, more scientifically accurate than not iterating over height.

If speed is an issue, either under normal usage, or in inner-loops, we could

1. Add more phi classes instead (less computationally-intensive than height)
2. Determine when to use height integration depending on the wind speed and column height


src/tephra2.ts
```diff
@@ -15,8 +15,9 @@ const kLargeDiskRadius = 10000;
 const kLargeDiskGridSize = 3000;          // larger = fewer calculations
 const kSmallDiskRadius = 500;
 const kSmallDiskGridSize = 300;
-const kLargeDiskMassThreshold = 1e11;      // when to use the large disk size (1e11 = VEI 4)
-const kNumSimulatedPhiClasses = 7;    // 7 or 15, or you'll need to pre-calculate more massFractions
+const kLargeDiskMassThreshold = 1e11;    // when to use the large disk size (1e11 = VEI 4)
+const kNumSimulatedPhiClasses = 8;       // 7 or 15, or you'll need to pre-calculate more massFractions
```
knowuh: Outdated comment: `7 or 15`

src/tephra2.ts
```diff
@@ -199,16 +211,24 @@ const tephraCalc3 = (
...
-    diskGrid.forEach(diskCell => {
-      const load = tephraLoadFromDiskCell(
-        x, y, diskCell.x, diskCell.y, windSpeed, phiCellMass, colHeight, settlingSpeed);
-      totalLoad += load;
+
+  let totalLoad = 0;
+
+  heights.forEach(height => {
+    simulatedPhiClasses.forEach(phi => {
+      // if (phi !== 3) return;
```
knowuh: Could remove this comment?

PR #100, opened 2020-02-07
Improve VEI calculations

sfentress:
This makes VEI an entirely computed value, removing from concern whether it is saved when an author or user modifies the value.

This removes `vei` and `stagingVei` from the simulations-store's properties, and instead makes it a computed value given mass and column height.

When the user sets VEI, this will simply set the mass and column height.

If the user requests the VEI, we will make our best guess as to the VEI given the mass and column height. If those properties were set using `setVEI`, our "guess" will always be correct. If the user manually sets
mass and height to some non-possible combination, we will return a value in-between the values suggested by the two properties.

This also adds `stagingVei` based on the staging mass and height values, so that the VEI slider updates as the user updates the mass and height sliders.

@knowuh I believe that this may work better than #98, because it should remove any of the odd behavior we've seen with the VEI in the past (VEI widget not showing the right value), and allow us to keep our defined set of authored values, while still allowing an author to set VEI in the authoring view.

This should fix this bug: https://www.pivotaltracker.com/story/show/169983751

> I am currently trying to author different set ups in the curriculum. In one set-up, I only have the wind direction and wind speed sliders showing. I wish to set the VEI (eruption volume and height) to a specific value--but still hide the slider. Whenever i set the value of VEI and then hide the slider, it does not hold the value I set after I hit (save authored state.

### DIFF
--- a/src/stores/simulation-store.test.ts
+++ b/src/stores/simulation-store.test.ts
@@ -1,0 +1,64 @@
+import { SimulationStore, SimulationModelType } from "./simulation-store";
+
+describe("simulation-store", () => {
+  let simulation: SimulationModelType;
+
+  beforeEach(() => {
+    simulation = SimulationStore.create({});
+  });
+
+  describe("vei", () => {
+    it("we can set vei and staging mass and column height will be set", () => {
+      simulation.setVEI(1);
+      expect(simulation.stagingMass).toBe(1e8);
+      expect(simulation.stagingColHeight).toBe(500);
+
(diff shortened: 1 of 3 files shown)


sfentress:
Note: this can be tested here: http://geocode-app.concord.org/branch/improve-vei/index.html

You can use the sliders to see how the VEI gets updates as mass and height are updated.

You should also be able to add this to a LARA authoring page to test authoring.

PR #105, opened 2020-02-11
Save data only after we change data for the first time

sfentress:
This should both fix the authoring issue where LARA reports that an author has changed the authoring data (and so throwing up the "Leave site?" dialog every time an author leaves the page), and will also stop
the student from saving ghost data, which can make loading authoring data and linked data more confusing.

Not saving the student data can also potentially improve Chris Lore's experience in https://www.pivotaltracker.com/story/show/170927758, where data was saved on the second page of a linked model. With these changes, the second model won't save data simply be being opened, but only after a real change is made.

We compare the initial state with what is being saved back. After we save once we'll always save, so as to avoid confusion when an author resets to initial conditions.

https://www.pivotaltracker.com/story/show/170648882

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -6,7 +6,7 @@ import * as iframePhone from "iframe-phone";
 
 import { AppComponent } from "./components/app";
 import { onSnapshot } from "mobx-state-tree";
-import { stores, serializeState, getSavableState, deserializeState, updateStores, SerializedState } from "./stores/stores";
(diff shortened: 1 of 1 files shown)


sfentress:
This can be tested by LARA authors by adding the branch http://geocode-app.concord.org/branch/improve-authoring/index.html as a model.

They should notice that they no longer get the "Leave site?" dialog when they try to leave a page without making changes, and that the "Save model" button only appears after they have made a real change.

PR #109, opened 2020-02-18
Reorganize and cleanup

sfentress:
This simply moves a bunch of components to subfolders, and deletes a couple old unused components.


sfentress:
:+1: on organizing the buttons. If you have good thoughts on those, feel free to make some changes when you come on!

PR #110, opened 2020-02-18
Add d3 charts

sfentress:
This adds initial scatter and wind direction/magnitude charts. I'm working now to try and connect them with actual wind data, but I realized I should get this part in first.

These can be demoed at http://geocode-app.concord.org/branch/add-d3-charts/index.html by opening the Data tab and then checking off "Show Demo Charts?" at the bottom of the authoring widget.

More styling work will need to be done, and more work to support dates and axes units as the wind work comes in.


sfentress:
Note that this PR contains three commits at the start that belong to #109. They'll disappear when that is merged in and this is rebased.

src/components/charts/canvas-d3-radial-chart.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const magScale = d3.scaleLinear()
+        .domain([0, maxMagnitude])
+        .range([ 0, maxRadius]);
+
+    // add axes. Assume NESW for now, can add props to customize later if needed
+    svgAxes.append("g")
+      .call(d3.axisTop(xScale).tickValues([0]).tickFormat(() => "N"));
+    svgAxes.append("g")
+      .attr("transform", "translate(" + chartWidth + ", 0)")
+      .call(d3.axisRight(yScale).tickValues([0]).tickFormat(() => "W"));
```
knowuh: I think this should probably be "E" for east.

src/components/charts/canvas-d3-radial-chart.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const lineWidthShrink = Math.max(0, Math.min(1, (data.length - 1000) / 2000));
+    ctx.lineWidth = 0.5 - (0.3 * lineWidthShrink);
+    ctx.strokeStyle = "#3c7769";
+    ctx.beginPath();
+    data.forEach(point => {
+      const px = center.x + Math.cos((90 - point[0]) * Math.PI / 180) * magScale(point[1]);
+      const py = center.y - Math.sin((90 - point[0]) * Math.PI / 180) * magScale(point[1]);
+      // const px = center.x + xScale(x);
+      // const py = yScale(y);
+      const headlen = 10; // length of head in pixels
```
knowuh: Thanks for doing the arrows. I realize those appear in all of our specs, but maybe ultimately we will just do small dots.  I am sorry that wasn't communicated better.  But also, I think having arrows might be cool for us in our discussions.

src/stores/charts-store.ts
```diff
@@ -0,0 +1,22 @@
+import { types } from "mobx-state-tree";
+
+const Chart = types.model("Chart", {
+  type: types.union(types.literal("scatter"), types.literal("radial")),
+  data: types.array(types.array(types.number)),      // (x,y) or (deg,mag) tuples: [[x,y], [x,y], ...]
```
knowuh: At some point we might want to also include a date attribute. TBD

src/components/charts/canvas-d3-radial-chart.tsx
sfentress: :+1: Fixed in the next PR.

src/stores/charts-store.ts
sfentress: Added in the next PR.

PR #116, opened 2020-02-21
Add wind data

sfentress:
This loads in the wind data from a .csv file, and adds a new data sampling utility class. It then updates the chart components to support Date objects, and puts more of the processing of the data into the Chart Store, to try and keep the code in the chart components as data-agnostic as possible.

(This second branch of add-wind-data was added because I had already told folks that the buttons would just be visible, which we don't want in master, and because I wanted to test removing a MST types issue in the other branch that turned out to be unnecessary.)


sfentress:
@mklewandowski Sorry for the delay, the radial scale and dots are now added and this PR is ready.

PR #117, opened 2020-02-24
Graph all wind data

sfentress:
This adds two simple blocks, "Graph Wind Data" and "All Wind Data" which fit together.

As part of this, this PR includes a big clean-up of lots of obsolete stuff in the blocks code, and in the simulations store, and extracts the blockly code-controlling code into its own controller. This could have been separated into another PR, but (1) this shows the motivation for the refactor (besides cleaner code, it makes it trivial to call another store from the blockly interpreter, which is needed for the graphing) and (2) the actually graphing code was tiny, given prior work.

Adding in the Sample and Filter blocks will be very simple, now that the Datasets controller's sample functions take in datasets and produce datasets, so already act as the kind of pipes we will need. The Sample block, for instance, is basically one line of code, besides the block definition, but it seemed to make sense to keep this initial work separate because of the blockly refactoring.


src/blockly/interpreter.ts
```diff
@@ -1,11 +1,16 @@
...
-// import { Interpreter } from "js-interpreter";
+import { BlocklyController } from "./blockly-controller";
 import { SimulationModelType } from "../stores/simulation-store";
 import { IBlocklyWorkspace } from "../interfaces";
+import { IStore } from "../stores/stores";
+import { Datasets, Dataset } from "../stores/data-sets";
 const Interpreter = require("js-interpreter");
 
-const makeInterperterFunc = (simulation: SimulationModelType, workspace: IBlocklyWorkspace) => {
+const makeInterperterFunc = (blocklyController: BlocklyController, store: IStore,
```
kswenson:
```suggestion
const makeInterpreterFunc = (blocklyController: BlocklyController, store: IStore,
```

sfentress: :+1:

PR #124, opened 2020-02-26
170909174 monte carlo histogram

mklewandowski:
This PR adds a histogram to the Monte Carlo that can be filled in by passing in a chart from our chart store.  A `HistogramPanel` component contains both the histogram and some accompanying text.  And a new `SvgD3HistogramChart` renders the actual histogram.  A few notes about things that will need to be modified as we move forward:
- at present there is a button to make a histogram.  It generate 300 random data points (from 0 to 400), adds them to a chart in the chart store, and then passes the chart to the histogram component to be displayed.  We will need to remove this fake data and pass in real data to the newly created chart.
- the threshold is currently set to a constant value of 201.  We will need to pass in a threshold defined in a block.
- at present the min and max for the tephra thickness values are set to 0 and 400 - these set the range of the x axis on the histogram.  Since the histogram will be built incrementally by stepping over a program, it is probably best to have a pre-defined tephra thickness range (otherwise if we pull the min/max values from the data, the histogram will render with different ranges each time we add data to the chart - this seems like it will be jarring to the user).  I'm not sure if 400 is a proper max, but this is what was shown in mock-ups
- I left in a button to toggle between a dot plot and a bar graph version of the histogram.  I think the bar graph is much better suited to our needs since we will incrementally add to the histogram and because the container div can be scaled along x or y (it will be much easier to display the bar graph in a predictable manner given the flexibility of the graph).

### DIFF
--- a/src/components/montecarlo/histogram-panel.tsx
+++ b/src/components/montecarlo/histogram-panel.tsx
@@ -67,16 +67,15 @@ export class HistogramPanel extends BaseComponent<IProps, IState>{
     const { width, height } = this.props;
     const { data } = chart;
 
-    const dataArray: any[] = data.map(d => d[0]);
     const threshold = kTephaThreshold;
     let above = 0;
-    dataArray.forEach(d => {
+    data.forEach((d: number) => {
       if (d > threshold) {
         above++;
       }
     });
-    const below = dataArray.length - above;
-    const abovePercent = Math.round(above / dataArray.length * 100);
+    const below = data.length - above;
+    const abovePercent = Math.round(above / data.length * 100);
     const belowPercent = 100 - abovePercent;
 
     return (
@@ -121,9 +120,9 @@ export class HistogramPanel extends BaseComponent<IProps, IState>{
   private addHistogram = () => {
       const numPoints = 300;
       const title = "Chart " + (this.stores.chartsStore.charts.length + 1);
-      const data: number[][] = [];
+      const data: number[] = [];
       for (let i = 0; i < numPoints; i++) {
-        data.push([Math.floor(this.randomG(3) * (kTephaMax + 1))]);
+        data.push(Math.floor(this.randomG(3) * (kTephaMax + 1)));
       }
       const xAxisLabel = "Tephra Thickness (mm)";
       const yAxisLabel = "Number of Runs";
--- a/src/components/montecarlo/histogram-panel.tsx
+++ b/src/components/montecarlo/histogram-panel.tsx
@@ -2,6 +2,9 @@ import * as React from "react";
 import styled from "styled-components";
 import { inject, observer } from "mobx-react";
 import { BaseComponent, IBaseProps } from "../base";
+import { HorizontalContainer, VerticalContainer } from "../styled-containers";
+import { SvgD3HistogramChart } from "../charts/svg-d3-histogram-chart";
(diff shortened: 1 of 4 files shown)


src/components/montecarlo/histogram-panel.tsx
```diff
@@ -32,19 +44,99 @@ export class HistogramPanel extends BaseComponent<IProps, IState>{
...
+  }
+
+  // TODO: for testing only, remove when we have real data
+  private addHistogram = () => {
+      const numPoints = 300;
+      const title = "Chart " + (this.stores.chartsStore.charts.length + 1);
+      const data: number[][] = [];
+      for (let i = 0; i < numPoints; i++) {
+        data.push([Math.floor(this.randomG(3) * (kTephaMax + 1))]);
+      }
```
sfentress: I don't quite understand the data structure. Why is it a 2d array single-length arrays? Why not use a 1d array?

mklewandowski: I was trying to be consistent with how the existing chart model structured data:
`types.array(types.array(types.union(types.number, types.Date)))`

mklewandowski: But it's probably better to expand the `data` type definition in the `charts-stores.ts`

mklewandowski: So we could do something like this:
`data: types.array(types.union(types.array(types.union(types.number, types.Date)),types.number))` in the `Chart` model in `charts-stores.ts`.  The data can be:
- an array of numbers
- an array of arrays of numbers or dates
This does break some other code so I believe we'll have to do some extra checking to ensure we have a 2d array in some cases.  Something like `Array.isArray(self.data[x])`.

mklewandowski: I added a commit that allows that data to also be stored as a simple array of numbers.  
Let me know what you think:
https://github.com/concord-consortium/geocode/pull/124/commits/1fdb0a842d6cdece7b071efb21a0d328ec14b67b

mklewandowski:
@sfentress I added 2 new commits.  One to store histogram data as a simple number array and one to improve the dot layout (and show the dots by default).  Take a look when you have a chance and let me know what you think.  Thanks!

sfentress:
I had totally forgotten that you needed to match the chart-store data structure. It makes total sense why it was like that now.

I'm not sure what's the most "correct" way to structure it going forward. I think your way makes sense, although it's a pity MST makes everything need to be so verbose -- in pure Typescript we'd compose a type and just say that `data` is this type, which would be more readable, but we definitely don't want to introduce a new MST model to handle points, because there'd be a lot of overhead.

The other way we could do it would be to have multiple `data` properties, like x-y data, date-y data, histogram data, etc.

I think let's keep this for now, and the next time we need to update this we should look at all our uses and (un)generalize.

The new dots look very nice. :+1:

mklewandowski:
@sfentress Yeah, I wasn't quite sure how to handle it either.  I didn't like adding a new property to hold the histogram points.  That seemed excessive and confusing.  I wanted to reuse the existing `data` property.  But I'm not sure if my original approach (store the values in an array with a single entry) or this new one (expand the type definition) is better.  I agree that we should circle back to this.  Especially if future changes result in difficulty accessing the `data` property because we have to add in extra and potentially confusing type checking.

PR #125, opened 2020-02-26
Small chart fixes

sfentress:
Small fixes for partial completion of https://www.pivotaltracker.com/story/show/171502604


sfentress:
Yes, those buttons are obsolete. I want to keep them around a little longer as they are the only place you can see the arrows in the radial chart, and I'm not clear if there's a final decision on them.

PR #126, opened 2020-03-03
170909187 risk level text

mklewandowski:
This PR completes the risk level text in the histogram panel by including the following changes:
- add a percentage complete to the histogram panel that is used to show a progress bar
- if the percentage is less than 100, do not show the risk level
- improve layout when we have no data
- minor refactor

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -389,6 +389,9 @@ export class AppComponent extends BaseComponent<IProps, IState> {
                   <HistogramPanel
                     width={ mapWidth }
                     height={ (height - 90) * .35 }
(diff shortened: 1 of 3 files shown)


mklewandowski:
@sfentress @knowuh the percent complete was still in the story specs so I built it for now knowing that we can turn it off pretty easily.

mklewandowski:
@sfentress @knowuh I made `percentComplete` in the `HistogramPanel` props optional so all we need to do to remove the progress bar (if we all agree that we do not want it) is not pass a `percentComplete` to the `HistogramPanel` in `app.tsx`.  This gives us the flexibility to go in either direction as we get more feedback.  I'm voting to merge this in now since removing the progress bar will now be a one line change.

sfentress:
:+1:

PR #132, opened 2020-03-05
Add risk warnings

mklewandowski:
When selected on the monte carlo tab, add a risk warning icon (displayed as a divIcon inside of a leaflet marker) for each item from the samples collection and display a risk level key.  Notes:
- at present all samples collections are shown and risk is computed with the default threshold of 201.  These values eventually need to come from the blocks.
- some useful functions and constants were moved to `monte-carlo.ts` as they need to be accessed from multiple locations

### DIFF
--- a/src/components/map/map-component.tsx
+++ b/src/components/map/map-component.tsx
@@ -4,7 +4,7 @@ import * as L from "leaflet";
 import { Map as LeafletMap, TileLayer, Marker, Popup, ScaleControl, Pane } from "react-leaflet";
 import "leaflet/dist/leaflet.css";
 import "../../css/map-component.css";
-import { iconVolcano, iconMarker, getCachedDivIcon } from "../icons";
+import { iconVolcano, iconMarker, getCachedDivIcon, riskIcon } from "../icons";
 
 import { CityType  } from "../../stores/simulation-store";
 import * as Scenarios from "../../assets/maps/scenarios.json";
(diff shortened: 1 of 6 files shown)


src/components/map/map-component.tsx
```diff
@@ -318,4 +326,23 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
+      riskLevel && riskItems.push(
+        <Marker
+          position={[samplesCollection.x, samplesCollection.y]}
+          icon={riskIcon(riskLevel.iconColor, riskLevel.iconText, true)}
+          key={"risk-" + i}
+        />
+      );
+    });
+    return riskItems;
+  }
```
mklewandowski: This function is where we will need to determine which samplesCollection actually have risk level icon shown on the map and which threshold to actually use (kTephraThreshold is just for testing).

src/components/montecarlo/histogram-panel.tsx
```diff
@@ -171,8 +152,11 @@ export class HistogramPanel extends BaseComponent<IProps, IState>{
       const numPoints = 300;
       const title = "Chart " + (this.stores.chartsStore.charts.length + 1);
       const data: number[] = [];
+      const shift = Math.random();
       for (let i = 0; i < numPoints; i++) {
-        data.push(Math.floor(this.randomG(3) * (kTephraMax + 1)));
+        let r = this.randomG(3) + shift;
+        if (r > 1) r = r - 1;
+        data.push(Math.floor(r * (kTephraMax + 1)));
```
mklewandowski: Simple shifting of the random histogram numbers to facilitate testing the different risk levels.  The rest of the changes in this module are just refactors.  Some code was moved to `monte-carlo.ts` for easier access.

PR #134, opened 2020-03-06
Use user-defined threshold from block when displaying histogram

mklewandowski:
This PR takes the threshold defined on the graph excedence block, stores it in the chart store, and uses it to display the threshold line on the histogram.  If no threshold is defined on the excedence block, then 0 is used.  Previously we were always showing a threshold of 201 on the histogram.

### DIFF
--- a/src/blockly-blocks/block-graph-data.js
+++ b/src/blockly-blocks/block-graph-data.js
@@ -101,7 +101,7 @@ Blockly.Blocks['graph_exceedance'] = {
 }
 Blockly.JavaScript['graph_exceedance'] = function (block) {
   var location = block.getFieldValue('locations')
-  var threshold = block.getFieldValue('threshold') || 0
+  var threshold = Blockly.JavaScript.valueToCode(block, 'threshold', Blockly.JavaScript.ORDER_ATOMIC) || 0;
 
   var code = `graphExceedance({location: "${location}", threshold: ${threshold}});\n`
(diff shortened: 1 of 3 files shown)


src/blockly-blocks/block-graph-data.js
```diff
@@ -101,7 +101,7 @@ Blockly.Blocks['graph_exceedance'] = {
 }
 Blockly.JavaScript['graph_exceedance'] = function (block) {
   var location = block.getFieldValue('locations')
-  var threshold = block.getFieldValue('threshold') || 0
+  var threshold = Blockly.JavaScript.valueToCode(block, 'threshold', Blockly.JavaScript.ORDER_ATOMIC) || 0;
```
sfentress: Ah, this is because I was torn over whether the threshold should be a block or a field, and so I ended up committing something that was half-and-half...

My personal preference is that all numbers that don't make sense to loop over really ought to be fields. That way there's no question of wrong types, default values are clearly visible, etc.

However, Noah's mockups have generally used blocks.

I think this is fine for now, and maybe we can remember to bring this up later.

mklewandowski: I agree - I would rather see this as a field.  It's much easier/faster to build a program this way, However, the specs do have a block so we should leave it this way for now.  Regardless, we should bring this up at some point.

PR #135, opened 2020-03-09
171419964 risk block

mklewandowski:
This PR adds a new "show risk" blockly block that allows a user to specify if the risk level for a given sample will be displayed on the monte carlo tab map with a user-defined threshold.  

One note, the risk level is added as an optional property on the `SamplesCollection`.  We could go an alternate route and create a new store that just holds risk levels, but this seemed like overkill and could create some linkage issues if samples collections are reset, deleted, etc.

### DIFF
--- a/src/blockly-blocks/block-risk-level.js
+++ b/src/blockly-blocks/block-risk-level.js
@@ -1,0 +1,34 @@
+Blockly.Blocks['show_risk'] = {
+  init: function() {
+    this.appendDummyInput()
+        .setAlign(Blockly.ALIGN_RIGHT)
+        .appendField("Show risk on graph for");
+    this.appendDummyInput()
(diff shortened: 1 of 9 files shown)


src/blockly-blocks/block-risk-level.js
```diff
@@ -0,0 +1,34 @@
+Blockly.Blocks['show_risk'] = {
+  init: function() {
+    this.appendDummyInput()
+        .setAlign(Blockly.ALIGN_RIGHT)
+        .appendField("Show risk on graph for");
```
sfentress:
```suggestion
        .appendField("Show risk on map for");
```

mklewandowski: Thanks for catching this!  Fixed!

PR #139, opened 2020-03-10
Add fast speed and speed slider

sfentress:
This adds a speed slider which can run the model much faster, while still showing the highlighting of the blocks (just flickering at the highest speeds) and the graphs/map update.

The top speed is about the fastest I was able to get it to go on my machine while still allowing the highlighting and React to update in real time.

To show the speed controls, check the box near the bottom of the authoring panel. We may eventually decide not to have all four options, but instead just two or three, but this is added so it can be tried out by the team.

I added an example Monte Carlo model in the Initial Code dropdown. On my machine, running the model for 200 steps takes:

````
Speed  |   Time
  0        4:30
  1        0:58
  2        0:20
  3        0:08
````
While for only 200 steps, the fastest speed seems too fast, for >1000 steps it seems rather more reasonable.


sfentress:
I'll bring this in even though I expect we'll be updating the slider, as at least that way we'll have the monte-carlo.xml initial code to use during the demo.

PR #143, opened 2020-03-11
Store current histogram tab and highlight risk icon on map

mklewandowski:
This PR stores the current histogram tab in the UI store, and then uses this to determine the name of the active histogram and its corresponding samples collection.  This is then used to determine if the risk shown on the map for a given samples collection should receive the blue active border.

This could result in a few odd outlier cases since it relies on the samples collections having unique names (although I couldn't break it when I tested).  The current programming structure is somewhat flexible in what a user can do so there might be a path to create samples collections with duplicate names.  For now, however, I feel the is adequate.  In the future, the the entire blockly ecosystem might benefit from more warnings, errors, or rules that prevent users from doing odd or malicious things.

### DIFF
--- a/src/components/map/map-component.tsx
+++ b/src/components/map/map-component.tsx
@@ -331,13 +331,17 @@ export class MapComponent extends BaseComponent<IProps, IState>{
     const { samplesCollectionsStore } = this.stores;
     const { volcanoLat, volcanoLng } = this.stores.simulation;
     const riskItems: React.ReactElement[] = [];
+    const tabIndex = this.stores.uiStore.currentHistogramTab;
+    const histogramCharts = this.stores.chartsStore.charts.filter(chart => chart.type === "histogram");
+    const currentChart = histogramCharts && histogramCharts[tabIndex];
+    const chartName = currentChart && currentChart.title ? currentChart.title : undefined;
     samplesCollectionsStore.samplesCollections.forEach( (samplesCollection: SamplesCollectionModelType, i) => {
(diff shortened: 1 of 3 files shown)


src/components/map/map-component.tsx
```diff
@@ -331,13 +331,17 @@ export class MapComponent extends BaseComponent<IProps, IState>{
     const { samplesCollectionsStore } = this.stores;
     const { volcanoLat, volcanoLng } = this.stores.simulation;
     const riskItems: React.ReactElement[] = [];
+    const tabIndex = this.stores.uiStore.currentHistogramTab;
+    const histogramCharts = this.stores.chartsStore.charts.filter(chart => chart.type === "histogram");
+    const currentChart = histogramCharts && histogramCharts[tabIndex];
```
kswenson:
```suggestion
    const currentChart = histogramCharts?.[tabIndex];
```

src/components/map/map-component.tsx
```diff
@@ -331,13 +331,17 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
+    const chartName = currentChart && currentChart.title ? currentChart.title : undefined;
```
kswenson:
```suggestion
    const chartName = currentChart?.title || undefined;
```

mklewandowski: I believe we need to move to typescript 3.7 to support optional chaining (GeoCode is on TS 3.6).  Which is a good reminder that we should look at updating GeoCode dependencies (not sure when that was done last).

PR #144, opened 2020-03-11
Improvements to the threshold histogram graph

mklewandowski:
This PR makes the following improvements to the threshold histogram plot:
- x axis maximum value is double the threshold
- improve y positions of dots by setting setting a higher y axis max when we have a small number of dots (previously in these cases the dots would have large vertical spacing between them instead of being drawn directly stacked on one another due to the y axis max)
- improve y positions of dots by using dot radius (rather than constant) to compute position offset from x axis line
- add small, single pixel x offset to dots so they can be more clearly distinguished from y axis line when dots are very small 
- update color to match other graphs and make dots more visible

### DIFF
--- a/src/components/charts/svg-d3-histogram-chart.tsx
+++ b/src/components/charts/svg-d3-histogram-chart.tsx
@@ -48,10 +48,15 @@ export const SvgD3HistogramChart = (props: IProps) => {
   // And apply this function to data to get the bins
   // Add all the cases exceeding the max into the max's bin
   const bins = histogram((data as number[]).map(d => Math.min(d, xDomain[1] - 1)));
-  let max = 0;
+  let binMax = 0;
   bins.forEach(bin => {
-    max = Math.max(max, bin.length);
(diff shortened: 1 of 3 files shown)


src/components/charts/svg-d3-histogram-chart.tsx
```diff
@@ -86,14 +86,14 @@ export const SvgD3HistogramChart = (props: IProps) => {
...
-    bins.forEach((bin, binIndex) => {
+    bins.forEach((bin) => {
       const binMap = bin.map((x, i) => i);
       svg.append("g")
         .selectAll("dot")
         .data(binMap)
         .enter()
         .append("circle")
-          .attr("cx", d => (1 + xScale(binIndex) * (chartMax / numBins) + dotRadius) )
+          .attr("cx", d => (1 + xScale(bin.x0 as number) + dotRadius) )
```
sfentress: Nice. That  makes a lot of sense. Pity the d3 docs didn't make it clearer.

PR #146, opened 2020-03-11
Force blockly svg resize if container is resized.

mklewandowski:
This PR fixes two issues:
- window is resized when blockly tab is not showing. When user returns to blockly tab, blockly div has wrong dimensions (0x0) and does not display content
- in certain cases where the browser screen changes size (such as using fullscreen button, showing/hiding dev tools, maximizing browser, etc.), the blockly div has wrong dimensions and does not take up full container space.

To fix, we monitor the container div size.  If it changes, call `Blockly.svgResize` which resizes the blockly container div to fit our parent div.  The only side effect that I've seen so far is that we might lose the pan when switching tabs, resizing, and returning to the blockly tab.

### DIFF
--- a/src/components/blockly-container.tsx
+++ b/src/components/blockly-container.tsx
@@ -13,7 +13,10 @@ interface IProps {
   height: number;
 }
 
-interface IState {}
+interface IState {
+  containerWidth: number;
+  containerHeight: number;
+}
 
 const Wrapper = styled.div`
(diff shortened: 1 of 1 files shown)


mklewandowski:
@sfentress I think it was here in the Blockly docs that gave me the big clue:
https://developers.google.com/blockly/guides/configure/web/resizable

PR #151, opened 2020-03-17
Add collection locations

sfentress:
This splits the create sample collections block into two blocks, requiring the user to create a location first.

It also updates block wording and adds a new version of the town label to the map.

Screen Shot 2020-03-17 at 1 08 40 PM


sfentress:
Good point on the first, updated.

For the second, I'm not sure. I think it's already pretty confusing that we have two maps, the only difference being that one of them hides the MC stuff -- I think it would make more sense to show everything on either map. But I think the issue might be compounded a bit if the locations showed up and the risk indicators didn't, because you might think the risk indicators were broken.

PR #155, opened 2020-03-20
Fix updating authoring state when menu changes

sfentress:
Previously, every time the authoring menu updated, we would load the application store with the properties from the authoring panel, and as a result we would throw all other state away.

This had the result (among other issues) that if an author updated a slider in the control panel and then hid the slider, the sliders would all reset to the initial values.

This fixes that problem by merging the current state of the application into the authoring panel props and loading the application from that merge.

[#169983751]

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -1,5 +1,6 @@
 import { inject, observer } from "mobx-react";
 import * as React from "react";
+const deepmerge = require("deepmerge");
 import { BaseComponent, IBaseProps } from "./base";
 import { MapComponent, Scenario } from "./map/map-component";
 import { LogComponent } from "./log-component";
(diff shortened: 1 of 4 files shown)


src/components/app.tsx
```diff
@@ -547,6 +549,19 @@ export class AppComponent extends BaseComponent<IProps, IState> {
...
 
+  private updateAuthoring = (authorMenuState: IStoreish) => {
+    // first get the state from the entire app, including slider values etc
+    const localState = serializeState(getSavableState()).state;
+
+    // delete the initialXml code that was serialized, or we will never update the blocks when
+    // the author changes the initial code
+    delete localState.simulation.initialXmlCode;
+
+    // the the authored state from the authoring menu overwrites local state
```
kswenson:
```suggestion
    // the authored state from the authoring menu overwrites local state
```

PR #164, opened 2020-04-28
Use pre-release version of Shutterbug to support snapshots in GeoCode

knowuh:
This should fix snapshots that are not working in the GeoCode module:

[#172483323]
https://www.pivotaltracker.com/story/show/172483323


sfentress:
I'm closing this for now, to get it out of the open PRs list, but @knowuh if this is still relevant, let us know.

PR #166, opened 2020-08-06
Wind direction fixes

mklewandowski:
The sample wind data is stored using direction that the wind is blowing to, but we use the direction the wind is blowing from when sampling a value and using it in the blockly blocks.  This resulted in the following issues that have been fixed in this PRR:
-incorrect wind direction in monte carlo simulation
-incorrect wind direction on radial plot when using filter block
-incorrect wind direction when using "Compute and visualize tephra with random wind sample" and the "Compute and visualize tephra with random wind sample and VEI" blocks

### DIFF
--- a/src/assets/blockly-authoring/code/monte-carlo-3-locs.xml
+++ b/src/assets/blockly-authoring/code/monte-carlo-3-locs.xml
@@ -9,8 +9,8 @@
             <field name="locations">City A</field>
             <next>
                <block type="create_sample_location" id="J1O+99A^f[4roxtt6Xw#">
-                  <field name="x">15</field>
-                  <field name="y">15</field>
+                  <field name="x">-10</field>
+                  <field name="y">-10</field>
                   <field name="name">Close-by</field>
                   <next>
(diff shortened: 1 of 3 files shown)


mklewandowski:
> I'm forgetting: would it not be possible to simply store the data as 'from' in the first place?

I actually tried that approach at first but it broke a bunch of things on the graphs where they do want the data used in the "from" format (for example, the scatter plots should use "from" when plotting the data) .  Sometimes we want it as "to" and sometimes as "from" depending on the context.  I'm not a fan of having these two uses in the app, but they want to move forward with it for now.

kswenson:
This has already been merged, but my question would be whether it would make sense to have a function rather than replicating the calculations. That way any future changes would be localized to a single function, e.g. if a decision is made to change the way the wind data are stored.

mklewandowski:
@kswenson yeah, I think this is a good point and ties to Sam's question of if we should just flip all of the wind data since there are competing notions of "to" and "from" used within the project when talking about direction and it's not obvious which should be thought of as the "standard" for our dataset.  I think this is a good thing to circle back to when someone digs back into this project.

PR #176, opened 2020-08-31
Display GPS Stations

sfentress:


src/components/map/map-triangulated-strain-layer.tsx
```diff
@@ -0,0 +1,265 @@
...
+            const lng = filteredData[i].longitude;
+
+            coords.push(lat);
+            coords.push(lng);
+            points.push([lat, lng, i]);
+
+        }
+
+        // Delaunator takes in a 1D array of coordinates organized [x1, y1, x2, y2, ...]
+        // It outputs a 2D array containing sets of vertecies
```
kswenson, line 199:
```suggestion
        // It outputs a 2D array containing sets of vertices
```

src/components/map/map-component.tsx
```diff
@@ -260,55 +262,89 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
-              p2Lat={crossPoint2Lat}
-              p1Lng={crossPoint1Lng}
-              p2Lng={crossPoint2Lng}
-            /> }
-            {isSelectingRuler && <RulerDrawLayer
-              map={this.state.mapLeafletRef}
-            />}
           </Pane>
+          {
+            isTephraUnit &&
```
kswenson, line 267: Would it make sense to separate this into a `TephraMapComponent` and a `SeismicMapComponent` (or whatever the right name would be)?

sfentress: Possibly. I don't want to optimize that quite yet (the tephra unit adds two panes, and the seismic unit one pane respectively, so it's not a lot of code), and there's the possibility that it won't quite make sense, with different parts needing to interweave at different z-indices. Probably each of the panes could be its own component, though, which would simplify the view somewhat.

src/components/map/map-triangulated-strain-layer.tsx
```diff
@@ -0,0 +1,265 @@
...
+                    const dist = Math.sqrt(Math.pow(data[i].latitude - data[k].latitude, 2) +
+                                           Math.pow(data[i].longitude - data[k].longitude, 2));
+                    if (dist < 0.1) {
+                        removablePoints.add(data[i].id);
+                    }
+                }
+            }
+        }
+
+        // Output station id for all points removed from the mesh
```
dstrawberrygirl, line 177: Looks like some dead code here

src/components/map/map-triangulated-strain-layer.tsx
```diff
@@ -0,0 +1,265 @@
...
+        let strainMax: number = 0;
+
+        for (let i = 0; i < mesh.triangles.length; i += 3) {
+            const strainOutput: StrainOutput = strainCalc({data: [ filteredData[points[mesh.triangles[i]][2]],
+                filteredData[points[mesh.triangles[i + 1]][2]],
+                filteredData[points[mesh.triangles[i + 2]][2]],
+            ]});
+
+            const strain = strainOutput.secondInvariant;
+            // const strain = Math.log10(strainOutput.maxShearStrain);
```
dstrawberrygirl, line 214: Dead code?

src/components/map/map-triangulated-strain-layer.tsx
sfentress: I think you're right, but I'm pretty much leaving Saul's strain-layer-component as-is until we actually start working on it. I didn't have the cognitive space to try and understand it all yet.

PR #183, opened 2020-09-10
GPS location graphs

sfentress:
This adds GPS locations graphs, using the old code for Datasets and Charts as much as possible, even though they don't match quite cleanly with the old work.


dstrawberrygirl (approved):
Looks good to me - the light validation on the dates works quite nicely too, though we don't check whether start is before the end date (it's pretty obvious if they get it wrong).

PR #184, opened 2020-09-10
173850510 velocity vectors

dstrawberrygirl:
Velocity vectors are representative of relative movement based on velocity magnitude and some use of current map scale to ensure the arrows are visible (and we're ignoring the curvature of the Earth since our area is relatively small). The way this is calculated can probably be improved, and/or simplified, but it does work for now.

I used one of the stations as a base point for creating a scale that matched the actual arrow length and positioned that scale near the main map scale on the left. This may be revised following feedback from the team.


src/components/map/map-component.tsx
```diff
@@ -441,4 +443,25 @@ export class MapComponent extends BaseComponent<IProps, IState>{
...
     return riskItems;
   }
+
+  private getMapScale = () => {
+    if (this.map.current) {
+      const map = this.map.current.leafletElement;
+      // Get the y,x dimensions of the map
+      const y = map.getSize().y;
+      const x = map.getSize().x;
+      // calculate the distance the one side of the map to the other using the haversine formula
```
kswenson, line 453:
```suggestion
      // calculate the distance from one side of the map to the other using the haversine formula
```

src/components/map/map-gps-stations-layer.tsx
```diff
@@ -49,15 +53,80 @@ export class MapGPSStationsLayer extends BaseComponent<IProps, IState> {
...
+      const arrowScale = mapScale / mapScaleAdjust;
+
+      const startLatLng = new LatLng(stat.latitude, stat.longitude);
+      // calculate the magnitude of the two component velocity values
+      const velMag = Math.sqrt((stat.eastVelocity ** 2) + (stat.northVelocity ** 2));
+      // Adjust the scale of the magnitude to improve visibility
+      // Stations move at a rate of meters per year, which is a small number.
+      // so we multiply by the map scale
+      const magnitude = velMag * arrowScale;
+      const dir = Math.atan(stat.northVelocity / stat.eastVelocity);
```
sfentress, line 70: This works because the `northVelocity` is almost always positive for these stations, but it would give wrong answers if any happened to head South -- it would rotate it by 180º. I only recently learned about `atan2`, which gives correct directions in all four quadrants of the plain.

```
  x   y         Math.atan(y/x)      Math.atan2(y, x)
  1   1             Pi/4                   Pi/4
  1  -1            -Pi/4                  -Pi/4
 -1   1            -Pi/4                   3/4 Pi
 -1  -1             Pi/4                  -3/4 Pi
```
Changing this to `Math.atan2(stat.eastVelocity, stat.northVelocity)` shows at least one (maybe exactly one? That wayward one we noticed before?) going South.

src/components/map/map-gps-stations-layer.tsx
```diff
@@ -49,15 +53,80 @@ export class MapGPSStationsLayer extends BaseComponent<IProps, IState> {
...
+
+      // we're working over a small area, so ignoring curvature of the Earth
+      const endLat = startLatLng.lat + (magnitude * Math.cos(dir));
+      const endLng = startLatLng.lng + (magnitude * Math.sin(dir));
+      const endLatLng = new LatLng(endLat, endLng);
+
+      // arrowhead
```
sfentress, line 77: Totally an aesthetic question, but it looks a little odd to me that the bigger arrows sometimes have huge arrow heads, and that also the width of the arrow head depends on the angle of the arrow (the more northerly, the tighter the arrow head).

If you calculate the points from the `endLatLng` instead of the start, it's easier to have a fixed-size head (with a maximum size depending on the magnitude, so you don't get arrow heads bigger than the vector). And if you add and subtract a little from the `dir`, instead of multiplying, you get constant angles.

Here's just a proposal, but again it's just aesthetics and feel free to dismiss:
```
const arrowBaseMag = Math.min(0.2, magnitude * 0.4);

const a1Lat = endLatLng.lat + (arrowBaseMag * -Math.cos(dir - 0.3));
const a1Lng = endLatLng.lng + (arrowBaseMag * -Math.sin(dir - 0.3));
const a1LatLng = new LatLng(a1Lat, a1Lng);

const a2Lat = endLatLng.lat + (arrowBaseMag * -Math.cos(dir + 0.3));
const a2Lng = endLatLng.lng + (arrowBaseMag * -Math.sin(dir + 0.3));
const a2LatLng = new LatLng(a2Lat, a2Lng);
```

sfentress (commented):
This all looks good. I left a few comments, they can all be safely ignored if you don't agree with them, except for the one that makes a difference about a station going northeast instead of southwest.

...actually, looking further, it also tweaks a lot of the hairy-looking vectors that were at the northwest. See screenshot, which also includes proposed tweaked arrowheads.
atan
atan2

src/components/map/map-gps-stations-layer.tsx
```diff
@@ -49,15 +53,80 @@ export class MapGPSStationsLayer extends BaseComponent<IProps, IState> {
...
+      const stat = allGPSStations.find(s => s.id === "QHTP")!;
+      const velMag = Math.sqrt((stat.eastVelocity  ** 2) + (stat.northVelocity ** 2));
+
+      const magnitude = velMag * adjustedScale;
+      const dir = Math.atan(stat.northVelocity / stat.eastVelocity);
+      const startLatLng = new LatLng(stat.latitude, stat.longitude);
+      // get the end point for the arrow we use for this station
+      const endLat = stat.latitude + (magnitude * Math.cos(dir));
+      const endLng = stat.longitude + (magnitude * Math.sin(dir));
+      const endLatLng = new LatLng(endLat, endLng);
```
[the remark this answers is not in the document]
dstrawberrygirl: I left that as an echo of the previous code while testing (to make sure the arrows did actually match in length with the scale) but this is a much cleaner version. Thanks, will change!

src/components/map/map-gps-stations-layer.tsx
dstrawberrygirl: Looks great, will do!

src/components/map/map-gps-stations-layer.tsx
dstrawberrygirl: Arrowheads look much better - I'll add this in, thanks!

sfentress (approved):
Looks great :+1:

PR #186, opened 2020-09-15
Set up initial structure of deformation model

sfentress:
Creates a model that draws a grid that can animate between states.

There's a WIP commit that adds it to app.tsx, which I left in for now, but if it would be easier to merge it with your work I can rebase it out.

### DIFF
--- a/src/stores/seismic-simulation-store.ts
+++ b/src/stores/seismic-simulation-store.ts
@@ -13,6 +13,10 @@ export const SeismicSimulationStore = types
     scenario: "Seismic CA",
     visibleGPSStationIds: types.array(types.string),      // by id
     selectedGPSStationId: types.maybe(types.string),
+
+    deformationModelStep: 0,
+    deformationModelSpeed: 100,      // steps / second
+    deformationModelEndStep: 500,
   })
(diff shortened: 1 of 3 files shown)


src/stores/seismic-simulation-store.ts
```diff
@@ -23,10 +27,35 @@ export const SeismicSimulationStore = types
...
+  .actions((self) => ({
+    startDeformationModel() {
+      self.deformationModelStep = 0;
+
+      const startTime = window.performance.now();
+
+      const updateStep = () => {
+        const dt = (window.performance.now() - startTime) / 1000;   // seconds since start
+        const step = Math.floor(dt * self.deformationModelSpeed);
+        console.log(startTime, dt, step);
```
kswenson, line 49: Remove?

dstrawberrygirl:
I merged this code into the work on setting up the deformation tab and other framework pieces so that all the tests would be able to pass. This included adding a simple block in the new Deformation category that, when placed and run, will start the deformation simulation. I'll therefore close this PR and make a new one that includes @sfentress work on the underlying model.

PR #187, opened 2020-09-17
174654244 Deformation tab

dstrawberrygirl:
This merges work on the model from @sfentress with some framework pieces to include the tab and a simple block to run the model.

This PR and subsequent PRs should probably remain separate from the current Seismic work in Master while that is nearing release


src/stores/stores.ts
```diff
@@ -114,6 +116,7 @@ function getStoreSubstate(blocklyStoreProps: string[], tephraSimulationProps: st
       blocklyStore: pick(blocklyStoreProps)(blocklyStore),
       tephraSimulation: pick(tephraSimulationProps)(tephraSimulation),
       seismicSimulation: {},      // nothing to save yet
+      deformationSimulation: {},
```
sfentress, line 119: I don't think this needs to be here. These are serializations of the various MST stores, and we don't have a `deformationSimulation` store yet.

src/stores/stores.ts
```diff
@@ -145,7 +148,10 @@ export const deserializeState = (serializedState: SerializedState): IStoreish =>
   if (serializedState.version === 1) {
     return serializedState.state;
   }
-  return {unit: {name: "Tephra"}, blocklyStore: {}, tephraSimulation: {}, seismicSimulation: {}, uiStore: {}};
+  return {
+    unit: { name: "Tephra" },
+    blocklyStore: {}, tephraSimulation: {}, seismicSimulation: {}, deformationSimulation: {}, uiStore: {}
+  };
```
sfentress, line 154: likewise.

sfentress (changes requested):
This looks good, except for the question about the serialized state.

src/stores/stores.ts
dstrawberrygirl: Good catch - I had rolled back changes after our discussions earlier in the week and moved things to be part of the seismic simulation instead, but these managed to sneak past me.

sfentress (approved):
Looks good :+1:

PR #188, opened 2020-09-17
174654780 Deformation blocks

dstrawberrygirl:
Building on the tab PR, this one adds the three blocks to create the strain simulation and to set velocity for both blocks.

I've used a placeholder in the store to capture the speed and direction values, I wasn't sure if the connections from the "create the sim" felt right, I don't have a good set of guidelines for that aspect (should it be above, to the left, etc?) so any feedback on this would be great.

### DIFF
--- a/src/assets/blockly-authoring/toolbox/seismic-toolbox.xml
+++ b/src/assets/blockly-authoring/toolbox/seismic-toolbox.xml
@@ -9,8 +9,6 @@
 
   <category name="Deformation Model" colour="15">
     <block type="deformation-create-sim"></block>
-    <block type="set_velocity_block_1"></block>
-    <block type="set_velocity_block_2"></block>
   </category>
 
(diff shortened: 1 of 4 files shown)


sfentress (commented):
Yeah, it's a little hard to know what's wanted, because the "Create simulation" block doesn't really have any meaning currently, besides "run the model," but it's basically out-of-order for that, and isn't enforcing any real order.

I think what would be most logical would actually be a single block with four values (v1, d, v2, d2). After all, you need all four of these values to run the model. You can't have fewer (without relying on defaults) and you can't have more (without overwriting earlier values).

If we need exactly four values, then it's really just busy-work to have the student drag in three different blocks. I feel like our instinct is to have more blocks because it feels more like "programming," but in this case it doesn't quite make sense.

(The code here is fine, but I would push back a little on the requirements.)

sfentress (approved):
:+1:

PR #189, opened 2020-10-01
174401018 Deformation model output

dstrawberrygirl:
Using the supplied formulae, the two plates only exhibit vertical movement relative to each other. If you use a deformation block and just set speeds to 20 for one block and -20 for the other you can see the slip between the plates. You have two options to create opposing movement - one is to only enter positive speeds and use opposing angles (0, 180) or set both angles to 0 and use + and - speeds. I can see that this might be a bit confusing for a student.

I've tried to include the horizontal components if the student enters angles that are not 0 or 180, but they don't have an effect at this time. I would welcome a second pair of eyes on all the math in here! The original formula uses atan, and not atan2, so I am not 100% sure the angular movement would work correctly without a rework there.


src/components/deformation/deformation-model.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import { inject, observer } from "mobx-react";
 import { BaseComponent } from "../base";
 import { IDisposer, onAction } from "mobx-state-tree";
+import { deg2rad } from "../../utilities/coordinateSpaceConversion";
```
kswenson, line 5: `deg2rad()` is imported but never used, while `deg2Rad()` is defined below and used throughout. 🤔

dstrawberrygirl: well spotted! I wasn't sure if VS code was leading me astray, I guess I shall remove the Rad version!

sfentress (approved):
I think it will be fine to bring this in as-is, since I'll probably be touching all the same code in the coming week.

PR #190, opened 2020-10-02
Update station filter

sfentress:
This adds speed and velocity filtering to the Show Stations filter, and also allows displaying a single GPS station.


src/stores/seismic-simulation-store.ts
```diff
@@ -16,9 +16,13 @@ export const SeismicSimulationStore = types
     showVelocityArrows: false
   })
   .actions((self) => ({
-    showGPSStations(stations: StationData[]) {
+    showGPSStations(stations: StationData[] | string) {
       self.visibleGPSStationIds.clear();
-      stations.forEach( stat => self.visibleGPSStationIds.push(stat.id) );
+      if (typeof stations === "string") {
+        self.visibleGPSStationIds.push(stations);
```
dstrawberrygirl, line 22: Do we want any validation here? a check to see if the available stations includes the entered station? Perhaps we could also have some simple regex to "starts with" the string that was entered? I am not sure if this aligns with the project needs,  but it wouldn't be much more of a reach maybe to allow for multiple stations if they are comma-separated?

dstrawberrygirl (approved):
I think this is good - only one optional suggestion for a bit more refinement on station names.

src/stores/seismic-simulation-store.ts
sfentress: Unsure. I think I wouldn't bother too much about validation for single strings -- I feel like either the station is there or it isn't. Comma separated wasn't requested, but it could be useful? It would require more explanation in the videos by Chris. We could check with him in the demo meeting.

PR #191, opened 2020-10-02
Show "Map" instead of "Conditions" for tab in Seismic unit

sfentress:
This adds this ability in a generic way, so it will be easy to add other custom names for units if needed.


dstrawberrygirl (approved):
Looks good to me!

PR #192, opened 2020-10-02
Add reset to Seismic store

sfentress:
This resets store props to their defaults (no stations visible or selected).

As the store evolves, we haver to remember to keep this up to date.


dstrawberrygirl (approved):
Looks good to me!

PR #193, opened 2020-10-05
Listen to state changes on unit and blocklyStore

sfentress:
This fixes the issue that making changes only on these stores (e.g. setting the unit to Seismic, then changing the toolbox and dragging out a block) wasn't getting saved.

This can be tested by a LARA Staging admin on this page: https://authoring.staging.concord.org/activities/20811/pages/308056/edit


dstrawberrygirl (approved):
I think this looks good!

PR #194, opened 2020-10-08
Prompt the student if the speed entered is too large

dstrawberrygirl:
This should also prevent the model from running until the student has corrected the speed.
This PR follows the work in https://github.com/concord-consortium/geocode/pull/189

### DIFF
--- a/src/blockly/interpreter.ts
+++ b/src/blockly/interpreter.ts
@@ -328,9 +328,7 @@ const makeInterpreterFunc = (blocklyController: BlocklyController, store: IStore
 
     addFunc("setPlateVelocity", (params: { plate: number, speed: number, direction: number }) => {
       if (Math.abs(params.speed) > Math.abs(seismicSimulation.deformMaxSpeed)) {
-        blocklyController.throwError(`Plate speed must be between ${-seismicSimulation.deformMaxSpeed} and ${seismicSimulation.deformMaxSpeed} mm/year`);
-        // tslint:disable-next-line:max-line-length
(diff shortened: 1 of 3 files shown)


src/blockly/interpreter.ts
```diff
@@ -322,11 +322,17 @@ const makeInterpreterFunc = (blocklyController: BlocklyController, store: IStore
...
       seismicSimulation.startDeformationModel();
     });
-    addFunc("setBlockVelocity", (params: { block: number, speed: number, direction: number }) => {
-      seismicSimulation.setBlockVelocity(params.block, params.speed, params.direction);
+
+    addFunc("setPlateVelocity", (params: { plate: number, speed: number, direction: number }) => {
+      if (Math.abs(params.speed) > Math.abs(seismicSimulation.deformMaxSpeed)) {
+        blocklyController.throwError(`Plate speed must be between ${-seismicSimulation.deformMaxSpeed} and ${seismicSimulation.deformMaxSpeed} mm/year`);
+        // tslint:disable-next-line:max-line-length
+        params.speed = Math.min(Math.max(params.speed, -seismicSimulation.deformMaxSpeed), seismicSimulation.deformMaxSpeed);
```
sfentress, line 333: I don't think setting the speed at this point does anything, does it? I would `return` after this to block make it clear it doesn't continue.

sfentress (approved):
Looks fine, but I'm a little confused about the speed getting set after the `throwError`.

src/blockly/interpreter.ts
dstrawberrygirl: Correct! thanks, I'll fix :)

PR #196, opened 2020-10-16
Finish deformation model math

sfentress:
This makes some fixes the the vertical movement (the version that's currently deployed always moves the same way no matter the inputs), and adds the horizontal movement to the deformation model.

As far as I can tell, the math is all correct. Rather than trying an algebraic integration to work out the horizontal displacement given the function for rate, I opted to simply calculate it iteratively -- working out the total horizontal displacement at year 1000 is found by summing all the displacements from year 0 to 1000. So as not to re-calculate the whole thing each time, the previous values are cached. I made the decision to keep all the previously-cached values (i.e. during a full run from 0 to 500,000 years, we might calculate once at 120 years, once at 250, once at 700, etc, depending on the randomness of `requestAnimationFrame`), which will slightly speed up any subsequent run at the same speed, for a small memory cost. I think this makes sense, but the savings for subsequent runs might be so minimal that it would make more sense just to keep the very last value.

There are a couple remaining UI odds-and-ends to the model -- labelling the plates, making the background grid clip to the triangle only on Run -- but I think those would make sense in another PR.

### DIFF
--- a/src/components/deformation/deformation-model.tsx
+++ b/src/components/deformation/deformation-model.tsx
@@ -39,6 +39,19 @@ const distanceScale = 50;
 // angle between the plates vertically (into the Earth)
 const dip = deg2rad(90);
 
+// to calculate the total horizontal displacement from an originX at year Y, we have to sum all the displacements
+// from year 0 to year Y. When Y > 100,000, this starts getting slow. Therefore we cache previous calculations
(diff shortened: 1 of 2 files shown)


dstrawberrygirl (approved):
This looks fantastic! 👍

PR #197, opened 2020-10-19
Update filter block with Lat Long corners and Historical Position checkbox

sfentress:


dstrawberrygirl (approved):
I think this looks good!

PR #198, opened 2020-10-19
Upate deformation model UI

sfentress:
Label blocks, move a couple things around so that everything fits in square model area (otherwise they can get cut off at some aspect ratios) and add a background grid that fades out as the model begins.


dstrawberrygirl (approved):
This looks great - the alpha fade is a really nice touch!

PR #199, opened 2020-10-20
Fix Lat-Lng tool labels extending beyond map bounds

sfentress:


dstrawberrygirl (approved):
A much better implementation - nice!

PR #200, opened 2020-10-20
Show strain rate labels

sfentress:
This adds labels showing the strain rate within Delaunay triangles. This only adds labels in triangles that are large enough given a specific zoom size to fit them (using some quick-and-dirty math -- we could be more exact by calculating the size of the largest circle we can enclose in the triangle, but this seems to work well enough).

This can be seen at http://geocode-app.concord.org/branch/show-strain-rate-labels/index.html, by adding the "calculate strain rate" and "show strain value" blocks, like in the screenshot.

Screen Shot 2020-10-20 at 11 31 11 AM

### DIFF
--- a/src/components/map/map-triangulated-strain-layer.tsx
+++ b/src/components/map/map-triangulated-strain-layer.tsx
@@ -7,16 +7,18 @@ import { inject, observer } from "mobx-react";
 import { BaseComponent } from "../base";
 import { StationData } from "../../strain";
 import "../../css/custom-leaflet-icons.css";
-import { LayerGroup, Polygon } from "react-leaflet";
+import { LayerGroup, Polygon, Marker } from "react-leaflet";
 import { equalIntervalStrainRanges, logarithmicStrainRanges } from "./map-strain-legend";
 import { ColorMethod } from "../../stores/seismic-simulation-store";
+import { divIcon } from "leaflet";
+import { strainLabelIcon } from "../icons";
 
 interface IProps {
   map: Leaflet.Map | null;
 }
 
 interface IState {
-  data: StationData[];
+  zoomLevel: number;
 }
 
 @inject("stores")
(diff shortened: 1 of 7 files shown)


src/components/map/map-triangulated-strain-layer.tsx
```diff
@@ -50,53 +63,50 @@ export class MapTriangulatedStrainLayer extends BaseComponent<IProps, IState> {
...
+      const perim2 = Math.sqrt(((p2.lat - p3.lat) ** 2) + ((p2.lng - p3.lng) ** 2));
+      const perim3 = Math.sqrt(((p3.lat - p1.lat) ** 2) + ((p3.lng - p1.lng) ** 2));
+      const perimeter = perim1 + perim2 + perim3;
+      const centLat = ((p1.lat * perim2) + (p2.lat * perim3) + p3.lat * perim1) / perimeter;
+      const centLng = ((p1.lng * perim2) + (p2.lng * perim3) + p3.lng * perim1) / perimeter;
+      const incenter = Leaflet.latLng(centLat, centLng);
+      const minPerimSize = 1.5 - ((zoomLevel - 6) * 0.4);
+      const largeSide = zoomLevel > 7 ? perimeter * 0.48 : perimeter * 0.45;
+      const smallSide = zoomLevel > 7 ? perimeter * 0.1 : perimeter * 0.25;
+      console.log(minPerimSize);
```
Blarfnip, line 97: I'm not 100% sure on the stance of leaving console.logs in when merging to master for this codebase. But I thought I'd point this out.

sfentress: :+1:

sfentress:
D'oh, stupid error, good catch. Thanks.

sfentress:
Actually, though I said in Slack that this was expected behavior for how we reset values in the stores from Blockly, I realized that this also masks a different problem, which is that you can put the "Render" block before the "Compute" block, and the team explicitly didn't want this. Fixed this by resetting the render flags when bounds are set, forcing the user to explicitly render after the bounds are set, and this will also fix the issue you noted.

Blarfnip (approved):
Perfect! This fix might also be necessary for the "Show GPS stations" block as well.

sfentress:
> Perfect! This fix might also be necessary for the "Show GPS stations" block as well.

Unless I'm misunderstanding how you're proposing this, that one would require how we reset (or don't) between runs.

Currently, running a program, then deleting all the blocks, and running again, will result in nothing changing in the state of the model. The app has behaved like this since Tephra, and if we want to change that we'll need to be explicit about the implications.

In this case, we can reasonably make the argument that the process of re-calculating strain will unset the render state, because of the requirement that you must render after calculating. But if you didn't have the calculate-strain block, the render state wouldn't change without an explicit Reset by the user.

In the case of the "Show GPS Stations," I don't think there's a block that can play that implicit resetting role.

Blarfnip:
> > Perfect! This fix might also be necessary for the "Show GPS stations" block as well.
> 
> Unless I'm misunderstanding how you're proposing this, that one would require changing how we reset the application state (or don't) between runs.
> 
> Currently, running a program, then deleting all the blocks, and running again, will result in nothing changing in the state of the model. The app has behaved like this since Tephra, and if we want to change that behavior we'll need to be explicit about the implications.
> 
> In this case, we can reasonably make the argument that the process of re-calculating strain will unset the render state, because of the requirement that you must render after calculating. But if you didn't have the calculate-strain block, the render state wouldn't change without an explicit Reset by the user, because there would be no block being run that would unset those values.
> 
> In the case of the "Show GPS Stations," I don't think there's a block that can play that implicit resetting role.

That totally makes sense. I think I was approaching from a programmer's mindset of the application's memory and state being blank at the beginning of any given run. 

When I think of it your way, with implicit resetting being reliant on another block that would change the state without drawing, then I completely agree with what is currently implemented.

PR #202, opened 2020-10-23
175096158 remove ruler from strain unit

Blarfnip:
Just a small change to remove the Ruler from the seismic unit. And a handful of rebases to catch it up to master.


sfentress (approved):
Looks great.

In the future, for a pull request, I'd recommend rebasing to clean up the history -- somehow there are two of the same commit here, and two merges from different branches. For a simple PR like this it's easy to ignore the noise, but for larger PRs it's very useful if the commit history forms a clear narrative, and doesn't tangle the branch network too much, so I think it's a good habit to have.

Anyway, this looks good. :+1:

PR #207, opened 2020-11-30
Monte-Carlo Risk Diamonds Toggle

Blarfnip:
Added a toggle in the authoring UI to toggle the risk diamonds in the location tabs during a Monte-Carlo test.

### DIFF
--- a/cypress/integration/branch/monte-carlo-panel.test.js
+++ b/cypress/integration/branch/monte-carlo-panel.test.js
@@ -17,6 +17,7 @@ before(() => {
     modelOptions.getModelOptionsMenu().click();
     modelOptions.selectInitialCode('Monte Carlo (3 locs)');
     modelOptions.getShowSpeedControl().click();
+    modelOptions.getShowRiskDiamondsOption().click();
     modelOptions.getModelOptionsMenu().click();
     blocksTab.setSpeedControl("fast")
     blocksTab.runProgram();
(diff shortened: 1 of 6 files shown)


sfentress (commented):
Hi Saul, the code looks good, but it looks like there's a failing test that will prevent this from building.

Blarfnip:
Hey @sfentress ,
I'm pretty unfamiliar with the testing pipeline in place for this project. I've removed the related tests, as I'm a little unsure of the proper testing procedure for an authoring toggle like this feature. 

As far as I know, all tests are passing now, but the Travis builds are somehow failing even though the logs are saying that all the tests are passing.

If you have any idea what's going wrong and could let me know, that would be great.

sfentress:
Hi Saul,

So it looks like it's just a linting issue. You can see this in two ways: 

1. In the build summary in this PR (where it says "Some checks were not successful) you can click Details next to the Travis builds, and then click the link next to where it says "The build failed." This will show you the console output, and the error is near the top.
2. Simply by running `npm run build` on the command line. You should get the same error.

In this case, it looks like the issue was just trailing whitespace, which is enforced by ts-lint. This is an annoyance, but if you use VSCode, you can open your User settings and add `"files.trimTrailingWhitespace": true`, which will mean you'll never have to worry about it in the future. Sublime also has a similar setting, as I expect other IDEs do.

I would re-add the tests, because the tests should be ok,* and then confirm that both `npm run build` and `npm run test` work from the command line.

Sorry it's such an annoyance adding in a simple change.

*Actually, the tephra monte carlo test is indeed flaky, because it involves random chance. If you wanted you could look into making it more robust, but if not, just leave it, and if it fails on Travis we can trigger the build again manually.

Blarfnip:
Apologies for the delay in fixing this. I just took another look at it and got it all working. I removed the whitespace and adjusted the cypress tests to work with the new authoring option. It's still flaky, hence the still failing build. But it passed all the tests locally on my computer.

sfentress (approved):
Looks great. :+1:

I will merge this in.

PR #208, opened 2020-12-29
Small fixes for rotated wind direction

sfentress:
This fixes two small left-over issues that I found after the work over the summer to rotate the wind 180º. I don't know if anyone else has noticed them yet, but I kept running into them when trying to fix the Monte Carlo bug (and for a while I was chasing a red herring thinking that the issue was related to rotating the wind).

This fixes:
* Though filtering a specific number for wind (e.g. direction=90º) correctly gives us wind FROM 90º, putting a Range block in the direction slot gave us wind TO.
* If we used either of the "Erupt with a random wind sample from" blocks, the block would modify the actual underlying data point, meaning it would disappear from the filter the second time we tried to find it (even after a reset).


mklewandowski (approved):
👍 LGTM!

PR #209, opened 2020-12-29
Fix mirrored tephra calculations

sfentress:
This fixes a tricky problem that Jie noticed, that running the Monte Carlo simulation seemed to give quite different results from running a model "by hand" and counting each time a location was covered in Tephra.

Screen Shot 2020-12-29 at 2 17 05 PM

After some research, I found that the problem seemed quite bizarre. Calculating the tephra thickness at a point due South of the volcano, when the wind was blowing South, would show no tephra, even though the map, which was using the exact same calculation, would show tephra to the South. But what was even weirder was that the calculation that the map was making *also* didn't show any tephra to the South, instead it showed Tephra to the West, even though the wind was blowing South and the map contours clearly showed Tephra to the South.

I finally figured out that what happened was

1. An small error was added when the new contour version of the Tephra layer was added in #34, flipping lat and long 
2. Noticing that the results seemed "mirrored" (reflected on the x=y axis) a change was added to mirror the underlying tephra calculations (4ff835f15), such that the double-reflection made the tephra layer look correct, even though it was calculating things in the wrong direction

I'm surprised that we never noticed that the Monte Carlo simulations were giving quite wrong answers, though I think perhaps we did and always put it down to random chance. Using an extremely-tightly bounded filter, like direction=4 (which only has a single sample) made this clearer (though it helps to use the `fix-wind-direction-for-range-filter` branch so you don't get the modifying-sample bug).

### DIFF
--- a/cypress/integration/branch/authoring.test.js
+++ b/cypress/integration/branch/authoring.test.js
@@ -58,13 +58,13 @@ context ('Authoring Options',()=>{
             leftPanel.getControlsTab().click();
             controlsTab.getEruptButton().click();
             //verify tephra is visible at a location south of volcano - brittle if the viewport size changes
-            map.getTephra().last().attribute('d').should('contain',"M237 242L249 254L249 336L243 342L243 354L237 361L237 367L234 370L228 370L222 376L201 354L201 336L195 329L195 254L210 239L234 239z")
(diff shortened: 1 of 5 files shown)


mklewandowski (approved):
👍 Good catch on this and thanks for the detailed explanation.  Glad to see these fixes to the Monte Carlo content!

PR #210, opened 2021-01-04
Add migration to convert author/student state to latest

sfentress:
When we created the Seismic work, I updated the shape of the student data, as we were modifying the MST stores. Because of this, old authoring or student data saved before this would not load in the new app. For this reason, the code soft-forked, with `production` being reserved for Tephra and `master` unable to merge directly into production.

This creates a framework for a series of migrations that can convert any old state to the latest version. At the moment, this only converts from the pre-seismic-unit states (which is what's saved, as of this commit, on production) to the post-seismic-unit state.

This belatedly bumps the serialization version number to 2. Because post-seismic-unit work may have been saved under version number 1, the migration duck-types to check if it needs migration.

The result of this work is that we will be able to

1. Release master as-is to production, instead of needing to cherry-pick tephra-specific features to production
2. Use production urls for the seismic work
3. Make further changes to the data structure as necessary, by increasing the version number and adding a new migration

I have QA'd this by saving authoring and student data on production and ensuring that it shows up when the branch is changed, and Chris Lore is doing the same.

### DIFF
--- a/src/utilities/migrate-state.ts
+++ b/src/utilities/migrate-state.ts
@@ -7,26 +7,19 @@ type Migration = (oldState: UnmigratedSerializedState) => SerializedState;
 
 const migrate01to02: Migration = (oldState) => {
   if (oldState.version === 1 && !oldState.state.unit) {
-    // serailization with version = 1 and a unit specified is identical to version 2
+    // serialization with version = 1 and a unit specified is identical to version 2
     const {simulation, uiStore} = oldState.state;
+    const {toolbox, initialCodeTitle, initialXmlCode, ...tephraSimulation} = simulation;
     return {
       version: 2,
       state: {
         unit: { name: "Tephra" },
         blocklyStore: {
-          toolbox: simulation.toolbox,
-          initialCodeTitle: simulation. initialCodeTitle,
-          initialXmlCode: simulation.initialXmlCode
+          toolbox,
+          initialCodeTitle,
+          initialXmlCode
         },
-        tephraSimulation: {
-          requireEruption: simulation.requireEruption,
-          requirePainting: simulation.requirePainting,
-          scenario: simulation.scenario,
-          stagingWindSpeed: simulation.stagingWindSpeed,
-          stagingWindDirection: simulation.stagingWindDirection,
-          stagingMass: simulation.stagingMass,
-          stagingColHeight: simulation.stagingColHeight
-        },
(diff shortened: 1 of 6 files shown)


mklewandowski (approved):
👍 LGTM!  It will be nice to get the tephra and seismic units working on prod!

src/utilities/migrate-state.ts
```diff
@@ -0,0 +1,53 @@
+import { UnmigratedSerializedState, SerializedState } from "../stores/stores";
+
+export const CURRENT_SERIALIZATION_VERSION = 2;
+
+type PartialMigration = (oldState: UnmigratedSerializedState) => UnmigratedSerializedState;
+type Migration = (oldState: UnmigratedSerializedState) => SerializedState;
+
+const migrate01to02: Migration = (oldState) => {
+  if (oldState.version === 1 && !oldState.state.unit) {
+    // serailization with version = 1 and a unit specified is identical to version 2
```
kswenson, line 10:
```suggestion
    // serialization with version = 1 and a unit specified is identical to version 2
```

src/utilities/migrate-state.ts
```diff
@@ -0,0 +1,53 @@
...
+        },
+        tephraSimulation: {
+          requireEruption: simulation.requireEruption,
+          requirePainting: simulation.requirePainting,
+          scenario: simulation.scenario,
+          stagingWindSpeed: simulation.stagingWindSpeed,
+          stagingWindDirection: simulation.stagingWindDirection,
+          stagingMass: simulation.stagingMass,
+          stagingColHeight: simulation.stagingColHeight
+        },
```
kswenson, line 29: I'm not that familiar with the details of the serialization structure, but it seems like it should be possible to do something like the following to reduce the boilerplate copying of every property:
```typescript
    const {simulation, uiStore} = oldState.state;
    const {toolbox, initialCodeTitle, initialXmlCode, ...tephraSimulation} = simulation;
    return {
      ...
      blocklyStore: {toolbox, initialCodeTitle, initialXmlCode},
	  tephraSimulation,
      ...
```
Just a thought.

sfentress: That's true. It's a little less explicit if you want to see what the final serialization (or "serailization") looks like, but it does make it clearer that the main difference between the structures is the removal of those three props from the old `tephraSimulation`, so that's nicer.

PR #211, opened 2021-01-05
Monte Carlo block changes

sfentress:
This updates several of the blocks related to the Monte Carlo simulation, simplifying them.

These changes are described in these stories:

* https://www.pivotaltracker.com/story/show/176084473
* https://www.pivotaltracker.com/story/show/176084327

The result of this is that the canonical Monte Carlo simulation program now looks like this:

Screen Shot 2021-01-05 at 11 06 18 AM

instead of 

Screen Shot 2021-01-05 at 4 59 46 PM


mklewandowski (approved):
👍 This looks good - I like the simplifications.

PR #214, opened 2021-01-13
Fix issue caused by migrating null data

sfentress:
When an author first creates a model, it passes {} to the migration, which doesn't deal with that correctly. This was causing `updateStores` to throw an exception, preventing the store listeners from ever being added.

This fixes the issue by making the initial type of the data loaded from LARA explicit such that it may be {}, and passes back a valid but empty state if so.

(This essentially returns the handling of the empty state to what it was before the migration work was added. We had tested the migration work against existing models, but hadn't tried creating a new model.)


mklewandowski (approved):
👍

PR #216, opened 2021-01-14
Save state on play

sfentress:
This

1. Ensures that learner state does not save when the model first loads
2. But does save when a learner clicks Play for the first time

This updates the user data model to add a useless boolean `hasRunOnce`. We ensure that this can't get set by an author but only by a student.

Then there is a little additional trickiness to ensure that the `initialState`, which we use to compare against to see whether the user had dirtied state, really is the state that gets loaded before the student performs any actions. Previously, because of the way Blockly loaded its code, we would get a ghost action that looked like it dirtied state.

### DIFF
--- a/src/stores/blockly-store.ts
+++ b/src/stores/blockly-store.ts
@@ -25,6 +25,9 @@ export const BlocklyStore = types
           // see https://github.com/microsoft/TypeScript/issues/31663
           (self[key] as any) = data[key] as any;
         });
+        // after we have loaded, set the current cml code to the initial. This will ensure that our
+        // first saveable state is the same as our initial state.
+        self.xmlCode = self.initialXmlCode;
       },
       runClicked: () => {
         // this does nothing but set this flag, so that we save user data on first run
--- a/src/stores/blockly-store.ts
+++ b/src/stores/blockly-store.ts
@@ -25,7 +25,7 @@ export const BlocklyStore = types
           // see https://github.com/microsoft/TypeScript/issues/31663
           (self[key] as any) = data[key] as any;
         });
-        // after we have loaded, set the current cml code to the initial. This will ensure that our
+        // after we have loaded, set the current xml code to the initial. This will ensure that our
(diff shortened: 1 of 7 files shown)


src/stores/blockly-store.ts
```diff
@@ -24,6 +25,13 @@ export const BlocklyStore = types
           // see https://github.com/microsoft/TypeScript/issues/31663
           (self[key] as any) = data[key] as any;
         });
+        // after we have loaded, set the current cml code to the initial. This will ensure that our
```
kswenson, line 28: `cml`?

src/stores/stores.ts
```diff
@@ -126,9 +130,12 @@ function getStoreSubstate(blocklyStoreProps: string[], tephraSimulationProps: st
 // gets the current stores state in a version appropriate for the authoring menu
 export const getAuthorableSettings =
   getStoreSubstate(blocklyAuthorSettingsProps, tephraSimulationAuthorSettingsProps, uiAuthorSettingsProps);
-// gets the current store state to be saved by an author or student
-export const getSavableState =
+// gets the current store state to be saved by an author
+export const getSavableStateAuthor =
   getStoreSubstate(blocklyAuthorStateProps, tephraSimulationAuthorStateProps, uiAuthorSettingsProps);
+  // gets the current store state to be saved by an student (the above, plus anything like run state or tab state)
```
kswenson, line 136:
```suggestion
  // gets the current store state to be saved by a student (the above, plus anything like run state or tab state)
```

mklewandowski (approved):
👍 Looks good (note I read through code but didn't try to run locally).

src/stores/blockly-store.ts
sfentress: xml. Changed.

(cml was actually our old MW format, but I don't think my fingers were thinking of that.)

src/stores/stores.ts
sfentress: :+1:

PR #218, opened 2021-01-20
Set MC chart max to be at least 10 mm

sfentress:
Previously if the user picked a 1mm threshold, the chart looked bad as the 2+ bucket would stack up just to the right of the threshold line.


mklewandowski (approved):
👍

PR #220, opened 2021-02-12
New strain calculation

sfentress:
This updates the calculation for strain, and makes a few other tweaks regarding the color scale and filtering of strain triangles.


src/stores/seismic-simulation-store.ts
```diff
@@ -141,7 +141,35 @@ export const SeismicSimulationStore = types
       // Each vertex is returned as an index to an array of coordinates
       const mesh = new Delaunator(coords);
 
+      // filter trianges by removing those that have a vertex with an angle that is too small
```
kswenson, line 144:
```suggestion
      // filter triangles by removing those that have a vertex with an angle that is too small
```

dstrawberrygirl (approved):
I think this looks good, not sure if you wanted to remove the old equations in a separate commit later?

sfentress:
I was leaving commented-out all the calculations that lead to the unused `areaStrain`, since this is also an important calculation that may get used at some point, and I figured it was better to leave a few commented-out lines than to try and dig through the history to recreate them.

PR #221, opened 2021-02-26
update browser tab title

mklewandowski:
Fix the browser tab name.  Change from "Tephra Example" to "GeoCoder."


sfentress (approved):
Sorry for the delay. +1:

PR #222, opened 2021-03-02
176787959 lat long fixes

mklewandowski:
This PR adds N and W labels to the boxes that are displayed in the lat/long tool.

Testable here:
http://geocode-app.concord.org/branch/lat-long-fixes/index.html

image


sfentress (approved):
:+1:

PR #223, opened 2021-03-02
Improve monte carlo display on small view ports

mklewandowski:
This PR makes small CSS improvements to the monte carlo stats so they display properly on small viewports.  

Testable here:
http://geocode-app.concord.org/branch/monte-carlo-display-fix/index.html

Results on 875 x 560 view port (the top image is how the monte carlo panel previously looked, the bottom image is the improved version):
image


sfentress (approved):
:+1:

PR #224, opened 2021-03-02
Add uniform axis scaling for position over time graph

mklewandowski:
This PR adds a `uniformXYScale` flag to the chart store so that the x and y tickmarks on the GPS location over time graph can be scaled uniformly.  

NOTE: it's a little awkward that we pass in a `height` to the `SvgD3ScatterChart` component and then I override it if `uniformXYScale` is true.  However, it seemed a lot more cluttered to do the calculation elsewhere and try to pass in the actual final height.  I'm open to thoughts on this.  

Testable here:
http://geocode-app.concord.org/branch/uniform-axes/index.html

image

### DIFF
--- a/src/blockly/interpreter.ts
+++ b/src/blockly/interpreter.ts
@@ -333,7 +333,8 @@ const makeInterpreterFunc = (blocklyController: BlocklyController, store: IStore
         duration: timeRange.duration ? timeRange.duration : undefined,
       };
       const dataset = Datasets.getGPSPositionTimeData(station, validTimeRange);
-      chartsStore.addArbitraryChart(dataset, "East (mm)", "North (mm)", `${params.station} Position over Time`, true);
+      chartsStore.addArbitraryChart(dataset, "East (mm)", "North (mm)", `${params.station} Position over Time`, true,
+                                    true);
     });
 
     addFunc("computeStrainRate", (filter: Filter) => {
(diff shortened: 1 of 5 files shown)


sfentress:
Hmmm, I haven't delved too deeply into the code, but I still see different tick intervals and different scales on the two axes.

In the screenshot below, HARV has 100s on the X axis and 50s on the Y axis, and the Y axis is about 150 pixels tall and the X axis between -700 and 0 is 200 pixels long.

Screen Shot 2021-03-02 at 3 36 38 PM

It also seems different for different stations:

Screen Shot 2021-03-02 at 3 26 10 PM

mklewandowski:
@sfentress yeah, the more I play around I see that it's totally inconsistent and I get variations based on what I did previously, view port size, other graphs, etc.  I'll have to look more closely.  Very confused as to why we get all of this variation.

mklewandowski:
@sfentress ah, I only fixed the S3 scatter plot and not the Canvas version.  Thanks for catching this.  I'll implement similar changes to the canvas version as well.

sfentress:
Yeah, sorry about the code duplication in the canvas vs svg versions. That was an unfortunate decision on my part.

mklewandowski:
> Yeah, sorry about the code duplication in the canvas vs svg versions. That was an unfortunate decision on my part.

No worries.  I pushed a fix.  Building now.

mklewandowski:
@sfentress something weird is happening with the Cypress tests, but I think you should be able to retest this branch.

sfentress (approved):
Looks good to me. 👍

PR #225, opened 2021-03-03
Add left/right tab state to UI store, save/restore to and from LARA

mklewandowski:
This PR moves the left and right tab state management put of the `App` component and into the `uiStore`.  The current tab index is then saved to and restored from LARA.

Running here (not much to see other than the fact that we can test that the tabbing still functions as expected):
https://geocode-app.concord.org/branch/tab-authoring/index.html

Here is an example running in the activity player.  When authored, I saved with the code tab and the data tab as the current tabs. When loaded, the code tab is restored on the left and the data tab is restored on the right:
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F20939.json&page=page_308580&preview

### DIFF
--- a/src/assets/test/test-serialization.json
+++ b/src/assets/test/test-serialization.json
@@ -32,7 +32,9 @@
             "showSpeedControls": true,
             "showBarHistogram": false,
             "showLog": false,
-            "showDemoCharts": false
+            "showDemoCharts": false,
+            "leftTabIndex": 0,
+            "rightTabIndex": 0
           }
         }
       },
@@ -76,7 +78,9 @@
             "showBarHistogram": false,
             "showLog": false,
             "showDemoCharts": false,
-            "showRiskDiamonds": false
+            "showRiskDiamonds": false,
+            "leftTabIndex": 0,
+            "rightTabIndex": 0
           }
         }
       }
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -35,8 +35,6 @@ import { UnitNameType } from "../stores/unit-store";
 interface IProps extends IBaseProps {}
 
 interface IState {
-  tabIndex: number;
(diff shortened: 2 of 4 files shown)


sfentress (approved):
This looks great.

By the way, you can test this stand-alone without LARA as well. Author some tabs and select them, and click the "Save current state to local storage" button. Clicking "Load" after a refresh now shows the selected tabs. (Hopefully this makes testing easier for any future such stories.)

Note, this is probably expected and fine, but the student state of the tabs will overwrite the authored state, if the student uses the model and comes back. (If there was any reason this was a problem, you'd want a student-specific version of `uiAuthorSettingsProps`). I only mention this because these new properties are the first time that a student's actions can make a change to the uiStore, so it's the first time it would have come up. For the block store, for instance, we have a student-specific `blocklyStudentStateProps` because the props need to be different.

mklewandowski:
> This looks great.
> 
> By the way, you can test this stand-alone without LARA as well. Author some tabs and select them, and click the "Save current state to local storage" button. Clicking "Load" after a refresh now shows the selected tabs. (Hopefully this makes testing easier for any future such stories.)
> 
> Note, this is probably expected and fine, but the student state of the tabs will overwrite the authored state, if the student uses the model and comes back. (If there was any reason this was a problem, you'd want a student-specific version of `uiAuthorSettingsProps`). I only mention this because these new properties are the first time that a student's actions can make a change to the uiStore, so it's the first time it would have come up. For the block store, for instance, we have a student-specific `blocklyStudentStateProps` because the props need to be different.

I'll make a note in the story about the student vs. teacher state in case Chris or Kiley want this to behave in a different manner.  Thanks for the reminder about local storage!

PR #226, opened 2021-03-03
174846693 gridlines

mklewandowski:
This PR adds a `gridlines` flag to the chart store so that gridlines can be shown on the GPS location over time graph.

Can be tested here:
http://geocode-app.concord.org/branch/gridlines/index.html

gridlines


sfentress (approved):
:+1:

PR #227, opened 2021-03-04
Fix seismic sizing on Firefox

mklewandowski:
This PR fixes a sizing bug in Firefox.  The panel below the seismic map was being shown with an incorrect height which caused it to take up half the screen.  Give it a static height to eliminate the bug.

Bug here:
image


sfentress (approved):
:+1:

PR #228, opened 2021-03-08
Parse and update Blockly XML to reposition program blocks

mklewandowski:
This PR improves the locations of blocks when a program is loaded - ensuring that blocks are, when possible, not positioned outside the workspace area.  This is done by determining the upper-left corner of the bounding box of the union of ALL program blocks and using this to move the entire program content to the upper-left of the Blockly workspace (essentially top aligning and left justifying the program structure while keeping the relative positions intact).

Simply trying to move things in bounds didn't have great results. Only moving a block or blocks that had strayed too far to the right or bottom in-bounds still frequently looked quite bad (you might end up with a block stuck against the right edge or stuck against the bottom). Furthermore, at load time we don't have access to block sizes (which is exacerbated by the fact that most of our blocks are connected to or children of other blocks - in this case we have neither size nor position information). Not having size or child positions makes determining where to move something that is out of bounds to the right/bottom a guessing game (unless we bake in a bunch of size constants for every block in the app).

I think this approach has the advantage of not only improving the presentation of out of bounds programs/blocks but of also improving the presentation of ALL programs as we can now think of all programs being aligned and justified in a uniform way when presented to a student.

test here:
http://geocode-app.concord.org/branch/block-visibility/index.html

Easiest way to test
-Create program
-Select Save current state to local storage in Model Options
-Refresh browser
-Select Load state from local storage in Model Options

Pasted Image at 2021_03_04_19_16 pm

### DIFF
--- a/src/components/blockly-container.tsx
+++ b/src/components/blockly-container.tsx
@@ -133,6 +133,48 @@ export default class BlocklyContainer extends React.Component<IProps, IState> {
       codeString = await intialCodeResp.text();
     }
 
+    if (codeString) {
+      // convert string to XML
+      const parser = new DOMParser();
+      const xmlDoc = parser.parseFromString(codeString, "text/xml");
+      // get the blocks from the XML
+      const xmlBlocks = xmlDoc.getElementsByTagName("block");
+      if (xmlBlocks.length) {
+        // find min and max x,y of blocks
+        let minX;
+        let minY;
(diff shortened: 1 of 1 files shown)


sfentress (approved):
Very nice. The code is very simple and straightforward, but it must have taken a bunch of exploration to get that to work. :+1:

PR #231, opened 2021-03-09
Remove WMAP GPS station

mklewandowski:
Remove the WMAP GPS station that has odd vector information.

Can be tested here:
https://geocode-app.concord.org/branch/remove-wmap/index.html

Map without WMAP:
image


sfentress (approved):
:+1:

PR #234, opened 2021-03-10
177256960 add set point tool

mklewandowski:
This PR adds a set point tool to the seismic model.  Changes includes:
- rename lat-long tool to Set Region
- add Set Point button
- when in set point mode a single lat/long point can be added and moved around the map
- Set Point and Set Region only appear in Seismic unit
- points or region are cleared when the mode is exited

Can test here:
http://geocode-app.concord.org/branch/set-point/index.html

set-point

### DIFF
--- a/cypress/support/elements/Map.js
+++ b/cypress/support/elements/Map.js
@@ -11,9 +11,12 @@ class Map {
     getRulerButton(){
         return cy.get('[data-test=Ruler-button]');
     }
-    getLatlngButton(){
-      return cy.get('[data-test=Latlng-button]');
+    getLatlngRegionButton(){
+      return cy.get('[data-test=Latlng-region-button]');
     }
+    getLatlngPointButton(){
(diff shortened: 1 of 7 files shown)


sfentress (approved):
I think this is fine. One question is whether this could have been done with an additional property in `LatLngRegionDrawLayer`, since it seems that 90% of the component is the same. But either way is fine.

Only other question is whether the single point should really be labeled "Corner 1".

mklewandowski:
> I think this is fine. One question is whether this could have been done with an additional property in `LatLngRegionDrawLayer`, since it seems that 90% of the component is the same. But either way is fine.
> 
> Only other question is whether the single point should really be labeled "Corner 1".

Yeah, I was expecting them to diverge more so I started with 2 modules.  I'll make a note to try to refactor this back into a single module if time permits in the sprint.  I also fixed the label.  It shouldn't say "corner 1."  Thanks for catching that.

PR #235, opened 2021-03-10
add plate colors to deformation model

mklewandowski:
This PR adds colored plates to the deformation model.  Changes include:
- left and right plates are added to deformation
- plates fade out and are hidden when model is run (similar to gridlines)

Can test here:
http://geocode-app.concord.org/branch/plate-color/index.html

plate-colors


sfentress (approved):
:+1:

PR #237, opened 2021-03-12
174846688 plate motion widget

mklewandowski:
This PR adds plate motion widgets to the deformation to show the speed and direction of the plates.  To build these controls, I took the existing wind control widget and expanded it.  The previous control that appeared at the top of the deformation model have been removed.

Can be tested here:
http://geocode-app.concord.org/branch/plate-motion-widget/index.html

image


sfentress (approved):
The code all looks great, nice work making the speed direction component generic.

I assume it was the team's choice not to have the widgets overlap, which I can understand, but have you tried looking to see what the model will look like in LARA? Because the model space now shrinks significantly when it has less height, which is the case in LARA, it will probably be extremely small. (Though of course they can always go to full screen.)

Screen Shot 2021-03-12 at 4 52 54 PM

src/components/widgets/speed-direction-widget.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { WidgetPanelTypes, kWidgetPanelInfo, round } from "../../utilities/widget";
+
+interface ValueDividerProps {
+  backgroundColor?: string;
+}
+const ValueDivider = styled.div`
+  width: 1px;
+  height: 21px;
+  margin: 0 5px 0 5px;
+  background-color: ${(p: ValueDividerProps) => `${p.backgroundColor ? `${p.backgroundColor}` : "#FFDBAC"}`};
```
kswenson, line 18:
```suggestion
  background-color: ${(p: ValueDividerProps) => `${p.backgroundColor || "#FFDBAC"}`};
```
or `??` depending on how you feel about other falsy values.

src/components/widgets/speed-direction-widget.tsx
```diff
@@ -0,0 +1,187 @@
...
+  position: relative;
+`;
+
+interface RotateProps {
+  rotate?: number;
+  top?: number;
+}
+const RotateDiv = styled.div`
+  position: absolute;
+  top: ${(p: RotateProps) => `${p.top ? `${p.top}px` : "25px"}`};
```
kswenson, line 31:
```suggestion
  top: ${(p: RotateProps) => `${p.top ?? 25}px`};
```
`||` would work as well if your TypeScript version doesn't support `??`

src/components/widgets/speed-direction-widget.tsx
```diff
@@ -0,0 +1,187 @@
...
+  height: 10px;
+  width: 10px;
+  transform: ${(p: RotateProps) => `${p.rotate ? `rotate(${p.rotate}deg)` : "rotate(0deg)"}`};
```
kswenson, line 34:
```suggestion
  transform: ${(p: RotateProps) => `rotate(${p.rotate ?? 0}deg)`;
```
and so on for a bunch of other values below

mklewandowski: Fixed.

src/components/widgets/speed-direction-widget.tsx
mklewandowski: Fixed.

src/components/widgets/speed-direction-widget.tsx
mklewandowski: Fixed.

mklewandowski:
I made another pass to both improve some of the model margins so they are more in line with the spec and to improve the positions of the text labels on the model.  None of this does very much to address Sam's size concerns.  I posted on Slack for the Geocode team to offer feedback, but at some point it's probably worth merging this and moving on.

PR #239, opened 2021-03-18
175037328 gps location time graph

mklewandowski:
This PR updates the seismic GPS location vs. time graph to include the following changes based on the updated UI spec:
- updated point color using new color ranges
- updated legend
- updated background color, title font, margins and sizing

Can test here:
http://geocode-app.concord.org/branch/gps-location-time-graph/index.html

image


sfentress (approved):
Looks good.

PR #240, opened 2021-03-22
176787693 direction widget

mklewandowski:
This PR adds an Explore Direction widget that is accessed via a button on the map in the Seismic unit.  This widget displays a compass and angular readout that can be changed with a slider.

Can be tested here:
http://geocode-app.concord.org/branch/direction-widget/index.html

direction-tool

EDIT, the direction tool can now be moved within the confines of the app window.
draggable


sfentress (approved):
This looks great to me, nice. :+1:

PR #242, opened 2021-03-22
Protect against performing spread on non-array

mklewandowski:
In some cases `getGPSPositionTimeData` results in a non-array value in `dataSet`.  In this case, the spread operation throws an error.  This was happening when we clicked on a GPS location on the map.  This PR adds protection around spreading non array values.

Can be tested here:
http://geocode-app.concord.org/branch/fix-array-spread/index.html

### DIFF
--- a/src/stores/data-sets.ts
+++ b/src/stores/data-sets.ts
@@ -94,7 +94,7 @@ export const Datasets = {
 
   getGPSPositionTimeData(name: string, timeRange?: TimeRange) {
     const dataSet = (PositionTimeDataSets as any)[name] as Dataset;
-    let filteredDataSet = [...dataSet];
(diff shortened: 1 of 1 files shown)


sfentress (approved):
This works. I think it's fine to bring this in, but do you have a good idea why it's sometimes not an array?

sfentress (approved):
(That was meant to be an approval.)

mklewandowski:
> This works. I think it's fine to bring this in, but do you have a good idea why it's sometimes not an array?

The value is actually undefined.  This happens when we click on a GPS station.  We call `getGPSPositionTimeData` with the station id and then use the return value to get the installed and last updated dates.  However, with many of the ids, the dataset is undefined.  So when I tried to perform array manipulation on undefined, we crashed.

PR #243, opened 2021-03-22
Update button styling

mklewandowski:
This PR updates the map button styling to match the UI spec.  Changes include:
- add correct background, hover, and active colors
- add new icons to ruler, set point, set region
- add cursor state when hovering

Can be tested here:
http://geocode-app.concord.org/branch/button-styling/index.html


sfentress (approved):
:+1:

PR #244, opened 2021-03-23
Fix GPS direction filter

mklewandowski:
This PR fixes a but that broke the GPS station direction filter.  The change between direction to and direction from only applies to the wind data.  Pass the filter function an extra parameter to determine if the to/from adjustment should be applied.

Note, it feels a little clunky to have an extra field in params for this, but trying to add it inside params.filter got sticky fast.  I made it optional and decided to leave the current implementation as the base case.  This means the to/from adjustment happens by default and those who don't want it, can turn it off.

Can test here:
http://geocode-app.concord.org/branch/fix-gps-direction-filter/index.html

### DIFF
--- a/src/blockly-blocks/seismic/block-seismic-gps-stations.js
+++ b/src/blockly-blocks/seismic/block-seismic-gps-stations.js
@@ -165,7 +165,7 @@ Blockly.JavaScript['seismic_filter_data'] = function (block) {
   let filterObj = JSON.stringify(filter);
   filterObj = filterObj.replace(/"([^"]+)":/g, '$1:')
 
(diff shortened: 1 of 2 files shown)


sfentress (approved):
Nice. I'm confused because I thought this worked before, but there's been so much confusing about direction-to/from that I have no idea what state it might have been in when it was working.

mklewandowski:
> Nice. I'm confused because I thought this worked before, but there's been so much confusing about direction-to/from that I have no idea what state it might have been in when it was working.

I haven't followed as closely as you, but the to/from stuff has been very confusing.  Certainly appears that it's been broken since December (and is broken on master and production).  Anyway, we should continue to keep an eye on it moving forward.

PR #246, opened 2021-03-24
Update GPS station graph to use offset into data to choose colors

mklewandowski:
This PR adds an offset to the returned dataset in `getGPSPositionTimeData` which can be stored in the chart data store as a new `dataOffset` field.  This is needed when displaying GPS position over time points that include a min/max date range.  The point colors need to be based on the position in the original dataset not their position in the filtered dataset. - the offset is used to get the position in the original dataset  This is so, for example, point 4000 in the original dataset always displays with the same color even if it is point 0 in a filtered dataset.

Can be tested here:
http://geocode-app.concord.org/branch/gps-color-fixes/index.html

Pasted Image at 2021_03_23_18_55 pm

Pasted Image at 2021_03_23_18_57 pm


sfentress (approved):
This looks good to me. :+1:

PR #247, opened 2021-03-24
Update GPS position over time legend to use blocks

mklewandowski:
Update the GPS position over time graph legend to use individual colored blocks instead of gradients.

Can test here:
http://geocode-app.concord.org/branch/legend-blocks/index.html

image


sfentress (approved):
:+1:

PR #248, opened 2021-03-24
make tephra semi-transparent

mklewandowski:
This is a draft PR that attempts to use Michael's knocked out transparency rendering of the tephra thicknesses on the map.  The goal is to maintain consistent transparency across all thicknesses instead of layering one thickness area on top of another which results in area being relatively opaque.

From the UI spec:
image

Result:
image

The solution is very simple and is based on this example of cutting a polygon out of another polygon using GeoJSON:
https://gist.github.com/andrewharvey/971b9db1900a62efa7635f6a6d337ae7

For each tephra thickness (n), we cut out the shape(s) from the next thickness level (n + 1).  This is done by simply adding the (n + 1) shapes when drawing the (n) layer and allowing the GeoJSON fill method to do the rest.  This assumes that each tephra thickness shapes(s) is contained within the shape of the previous thickness.  This is where I am uncertain.  In practice this seems to be the case, but I'm not sure this is always true.  Will the (n + 1) shapes always be bounded by or contained within the (n) shapes?

Can test here:
http://geocode-app.concord.org/branch/semi-trans-tephra/index.html

### DIFF
--- a/src/components/map/map-tephra-thickness-layer.tsx
+++ b/src/components/map/map-tephra-thickness-layer.tsx
@@ -121,7 +121,10 @@ export class MapTephraThicknessLayer extends BaseComponent<IProps, IState> {
 
     private renderGeoJson(geojson: MultiPolygon[]) {
         let gradientIndex = 0;
-        return(geojson.map(multipolygon => {
+        return(geojson.map((multipolygon, i) => {
+            if (i < geojson.length - 1 && geojson[i + 1].coordinates.length > 0) {
+                geojson[i + 1].coordinates.forEach((item) => multipolygon.coordinates.push(item));
+            }
             return (<GeoJSON key={this.keyval++}
                 stroke={false}
                 fillOpacity={0.8}


sfentress (approved):
I think this looks good. 

I can see the concern, because of the way the the points get "cut out" we actually have the potential to create a new polygon instead of cutting one out (the multipolygon object can create disconnected polygons). However, after playing with it a bit I think it wouldn't be such a big deal even if it were possible.

If we did get a situation where a layer above were not fully enclosed by the layer below, we can get a slightly funky situation:

Here is a fake demo of a very irregular eruption, using manipulated data, where the top layer has been shifted a little over, (rendered using the original code):

Screen Shot 2021-03-25 at 9 25 43 AM

When we apply the cutout code, we can see that the part that is overlapping correctly got cut out, and so is semi-transparent, while the part that is out of bounds is now totally opaque:

Screen Shot 2021-03-25 at 9 26 09 AM

What's happening is that the layer below added the overlapping shape, which meant it cut out the overlap but actually added the overhang, so now that part has two layers and is opaque.

However, even this unlikely result doesn't in any way crash the system or result in shapes that are actually incorrect. And in the real model, even if some overlapping might happen it would be more on the order of a pixel or two than this big offset.

Finally I'm pretty sure that the tephra algorithm can't actually give us non-enclosing topography, because of the way that it's the result of an integration and I think an overhanging tephra layer would be akin to a graph with an overhang, i.e. not a function that could be produced.

PR #249, opened 2021-03-25
Content scroll

mklewandowski:
This PR adds a min width to the app and horizontal scrollbars if the viewport width is smaller than the min width.  Changes include:
- a few components were moved from the `MapComponent` into `OverlayControls` so they could be better positioned
- I fixed the tabs in `OverlayControls`  which is why the diff looks so big

Note: I set the min width to 900, but we might go a little smaller to avoid scrollbars in LARA/AP in responsive view.  I'm waiting for feedback to make a final decision which will go into this PR.  In the meantime, I think the rest of the code can be reviewed.

Can test here:
https://geocode-app.concord.org/branch/content-scroll/index.html

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -50,7 +50,7 @@ const App = styled.div`
     flex-direction: row;
     height: 100vh;
     background-color: #ffffff;
-    min-width: 1100px;
+    min-width: 800px;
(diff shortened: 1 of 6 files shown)


sfentress (approved):
Code looks fine, have not attempted to test functionality.

PR #250, opened 2021-03-25
Always show scrollbars to prevent block shake

mklewandowski:
This PR fixes the block shake bug that appears in GeoCoder.  The previous approach of using scrollbars did not work as LARA prevents scrollbars in iframes.  Instead, this solution is very simple.
  
The shaking occurred when the following happened:
- GeoCoder was displayed in a small view port such that the Blockly workspace was not wide enough to fit some of our larger blocks.  At the same time, we were hiding the Blockly toolbox was hidden (and as a side effect the trashcan and scrollbars were also hidden).
- a program was loaded that contained blocks which did not fit in the provided Blockly workspace.
- Blockly  would shift back and forth between having the left edge of the block against the left edge of the workspace and having the right edge of the block against the right edge of the workspace (the shaking).

Solution:
- specify that the blockly workspace always has scrollbars.  When the blocks are loaded, scrollbars are added and Blockly can position the block freely in any position.

See gifs for before and after.
shake
noshake

### DIFF
--- a/src/components/blockly-container.tsx
+++ b/src/components/blockly-container.tsx
@@ -101,7 +101,8 @@ export default class BlocklyContainer extends React.Component<IProps, IState> {
         startScale: 0.8,
         maxScale: 2,
         minScale: 0.2
-      }
+      },
+      scrollbars: true
     };
 
     if (hideToolbox) {


sfentress (approved):
I think this is fine, and nice work finding a solution.

One thing is that it looks a little odd that the viewport seems to start out in the lower right. Probably it was always like this, but now that we can see the scrollbars it makes it more obvious:

Screen Shot 2021-03-25 at 2 55 46 PM

But this is not a reason to hold back the PR.

mklewandowski:
@sfentress I think it was always doing that (one of the reasons the block positioning is so confusing).  Here is a screenshot from production (unless we're looking at different things).
image

PR #252, opened 2021-04-07
Update Cypress wait times

mklewandowski:
This PR makes a couple of small improvements to the Cypress test wait times:
- when running the data tab test wait an additional second for the program to run.  I believe this was the cause of some of the failures I saw in the previous weeks.  App elements were missing because the program run was incomplete.
- reduce the wait time when running the monte carlo tests.  The wait time was 15 seconds.  I reduced it to 4 seconds.  The 15 seconds appeared to be based on using the standard program run speed, but we set the speed to "fast" and do not need to wait 15 seconds.


eireland (approved):
👍

PR #256, opened 2021-08-24
Earthquakes (scaled)

sfentress:
This adds some scaling options to the model, such that the model author can pretend the width of the model is different from what it actually is, and the time scale is different.

One way to play with it is to set up a model such as in #255, run it, and then after it has run change

* The actual width of the model
* The apparent width of the mode
* The apparent year scaling

The first option will change the curvature of the lines and the shape of the triangle, as the actual simulation will have been run closer to or further from the fault line. The second two options will only change the scale label and the year label.

The year scaling will also come into play when users can step through the model year by year in code.

This also adds the ability to hide the year label altogether.

https://geocode-app.concord.org/branch/earthquakes-scaled/index.html

### DIFF
--- a/src/components/authoring-menu.tsx
+++ b/src/components/authoring-menu.tsx
@@ -90,6 +90,7 @@ const AuthoringMenu: React.SFC<IProps> = (props) => {
               min={0.01} max={100} step={0.01}/>
             <DatNumber path="seismicSimulation.deformationModelApparentYearScaling" label="Apparent year scale" key="deformationModelApparentYearScaling"
               min={0.0001} max={1} step={0.0001}/>
+            <DatBoolean path="seismicSimulation.deformationModelShowYear" label="Show years?" key="deformationModelShowYear" />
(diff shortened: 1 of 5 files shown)


mklewandowski:
@sfentress are we off by a factor of 10 in the apparent width that is in the authoring and displayed in the deformation model?
image

mklewandowski (changes requested):
Looks good.  I'm still trying to wrap my head around what exactly the model width authoring control does, but I think I'm starting to get it.  The only issue that I noticed is that the apparent width setting in the authoring doesn't seem to match the apparent width shown on the deformation model.

sfentress:
@mklewandowski The scale marker is like a map's scale marker: the distance is from one end of the line to the other. The line is a tenth of the width of the model, and shows a tenth of the apparent width you selected.

If I'm being dense and missing what you're saying, LMK. Otherwise I *think* this is fine, and will bring it in.

mklewandowski:
Got it.  My confusion.

PR #257, opened 2021-08-31
Deformation blocks

sfentress:
This adds the blockly code to set up a student-controlled deformation simulation with earthquakes.

This can be tested at https://geocode-app.concord.org/branch/deformation-blocks/index.html

To set this up:

1. Select unit Seismic and toolbox GPS & Earthquakes
2. Set the Apparent year scaling to 0.001 (you can type it in, though it sometimes takes a couple tries to put the decimal in the right place). This will make it so you see significant deformation at 500 years.
3. Set Earthquakes -> User
4. Add the speed control so the model runs faster. (If the speed control exists, the default starts at faster than if it doesn't exist).
5. Set up code as in the gif below:

### DIFF
--- a/src/components/deformation/deformation-model.tsx
+++ b/src/components/deformation/deformation-model.tsx
@@ -144,9 +144,8 @@ export class DeformationModel extends BaseComponent<IProps, {}> {
     const { deformationModelStep: year, deformationModelEarthquakesEnabled,
       deformationModelRainbowLines, deformationModelWidthKm,
       deformationModelApparentWidthKm, deformationModelApparentYearScaling,
-      deformationModelShowYear } = this.stores.seismicSimulation;
-    const vSpeed = this.getRelativeVerticalSpeed();     // mm/yr
-    const hSpeed = this.getRelativeHorizontalSpeed();
+      deformationModelShowYear, relativeVerticalSpeed: vSpeed,
+      relativeHorizontalSpeed: hSpeed } = this.stores.seismicSimulation;
 
     // plates
     if (year < this.fadeOutTime) {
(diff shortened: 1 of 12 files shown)


src/components/deformation/deformation-model.tsx
```diff
@@ -568,16 +547,39 @@ export class DeformationModel extends BaseComponent<IProps, {}> {
...
+      const yearsToEarthquake = Math.abs(maxDisplacement / speedInKmYr);
+      const count = Math.floor(year / yearsToEarthquake);
+      return {
+        count,
+        yearsSinceEarthquake: year % yearsToEarthquake,
+        distanceTravelledDueToEarthquakes: count * (maxDisplacement / 3)
+      };
+    } else if (deformationModelEarthquakeControl === "user") {
+      // for user-defined earthquakes, the model maintains the count of earthquakes and the year of the last one.
+      // we just need to calculate the distanced the plates had traveled at the last earthquake
```
mklewandowski, line 568: "distanced " -> "distance"

mklewandowski:
I must be missing something obvious, but when I run locally I cannot get the deformation model to do anything with the new blocks.  I've tried tweaking setting, refreshing... what am I missing?
no-def

mklewandowski:
> I must be missing something obvious, but when I run locally I cannot get the deformation model to do anything with the new blocks. I've tried tweaking setting, refreshing... what am I missing?

Ah, never mind, it's my apparent year scaling.  I think I got bit by typing it in and the autocomplete overwrote my change.  It's working.

mklewandowski (approved):
👍 LGTM!  We will need to make sure we convey to others how to configure the authoring settings, but I think it's working well.  Also note that the same issue related to the model width being off by a factor of 10 from the previous PR is here as well.

PR #258, opened 2021-09-07
improve deformation block input values

mklewandowski:
This PR includes small improvements to the deformation block:
- speed field can only be a number block
- direction field can take number or string
- direction field can have value "North" or "South" which are converted to 0 and 180.
- show error message if invalid string is used in the direction field

This PR can be tested here:
https://geocode-app.concord.org/branch/deformation-block-updates/index.html


sfentress (approved):
Looks good. :+1:

PR #259, opened 2021-09-07
add URL params to set unit and authoring controls visibility

mklewandowski:
This PR adds two URL parameters that can be used to configure the initial simulation state.
- `hide-model-options`: when included, the model options authoring dialog in the upper-right is hidden.
- `unit={name}`: when included we configure the application for the specified unit. Unit name can currently be set to `Seismic` or `Tephra` (e.g., `unit=Seismic`). Values are case sensitive.

This PR can be tested with the following URLs...

No URL params:
https://geocode-app.concord.org/branch/set-seismic-from-url/index.html

Hide the model options:
https://geocode-app.concord.org/branch/set-seismic-from-url/index.html?hide-model-options

Set the unit to Seismic:
https://geocode-app.concord.org/branch/set-seismic-from-url/index.html?unit=Seismic

Hide the model options AND set the unit to Seismic
https://geocode-app.concord.org/branch/set-seismic-from-url/index.html?hide-model-options&unit=Seismic

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 The goal of GeoCode is to get students exploring geological threats using
 the practice of programming. In this case, we are using Blockly to generate
-code that runs a simple tephra distribution model.
+code that runs a simple tephra distribution model and seismic activity model.
 
 Students change eruption parameters using [Blockly](https://developers.google.com/blockly/).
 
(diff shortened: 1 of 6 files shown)


sfentress (approved):
This is probably fine for the PT story as specified. If the authors are going to start asking for changes to these defaults, however, I think it would be much better to simply bring in generic url-parameter-authoring, so that the authors can construct their own urls from the authoring.

We actually had this a while ago. It was originally added here: https://github.com/concord-consortium/geocode/commit/65d62700, and removed here: https://github.com/concord-consortium/geocode/pull/73. I think it would be possible to integrate it into the current merging of props that happens during the loading of authored data.

This would let us return to the state of having a "Generate URL" button in the authoring menu.

However, again, that's overkill for this story, and would also result in ugly urls.

One small change I might suggest:

1. Also support `?unit=Tephra`, for symmetry
2. If a unit is specified in the url, automatically hide the authoring menu

This would allow you to have two smaller, obvious urls for the two units.

sfentress:
D'oh, you have number 1 already. I knew that.

mklewandowski:
> This is probably fine for the PT story as specified. If the authors are going to start asking for changes to these defaults, however, I think it would be much better to simply bring in generic url-parameter-authoring, so that the authors can construct their own urls from the authoring.
> 
> We actually had this a while ago. It was originally added here: [65d6270](https://github.com/concord-consortium/geocode/commit/65d62700), and removed here: #73. I think it would be possible to integrate it into the current merging of props that happens during the loading of authored data.
> 
> This would let us return to the state of having a "Generate URL" button in the authoring menu.
> 
> However, again, that's overkill for this story, and would also result in ugly urls.
> 
> One small change I might suggest:
> 
> 1. Also support `?unit=Tephra`, for symmetry
> 2. If a unit is specified in the url, automatically hide the authoring menu
> 
> This would allow you to have two smaller, obvious urls for the two units.

Yeah, I'm with you on this too.  If this gets any more complex, then we should just make it so someone can author their own URL for whatever authoring settings that they want.  It could be a button in the model options to make the URL (like you did in CBio).  I'll add change 2, merge, and investigate your suggestions.  There might end up being a little more GeoCode time to do something like this.

PR #260, opened 2021-09-08
Show error if plate speed is not in range on user block

mklewandowski:
This PR enforces a plate speed range (currently 0 - 50) when the user runs using an Increase Deformation by Deformation Rate block.  If a user specified value is outside the range, an error message is shown.

Can be tested here:
https://geocode-app.concord.org/branch/deformation-speed-range/index.html

image


sfentress (approved):
:+1:

PR #261, opened 2021-09-08
Unify and update block colors

mklewandowski:
This PR updates and unifies block colors in each category so that each category has blocks of the same color and the colors follow those seen in the UI spec.

Can be tested here:
https://geocode-app.concord.org/branch/block-colors/index.html

image


sfentress (approved):
:+1:

PR #262, opened 2021-09-08
176084553 remove block

mklewandowski:
This PR removes the Add Town block from the primary, in-use toolboxes.

Can be tested here:
https://geocode-app.concord.org/branch/remove-block/index.html


sfentress (approved):
:+1:

PR #263, opened 2021-09-08
Update the apparent year scale default value to .001

mklewandowski:
This PR changes the apparent year scale default value in the seismic unit from 1 to .001.  This value is used by default in the model and in the authoring model options dialog where the user can change the value.

Can be tested here:
https://geocode-app.concord.org/branch/apparent-year-scale/index.html

### DIFF
--- a/src/stores/seismic-simulation-store.ts
+++ b/src/stores/seismic-simulation-store.ts
@@ -51,7 +51,7 @@ export const SeismicSimulationStore = types
     deformationModelShowYear: true,
 
     // changes the visible value for years passed, and when students step through years manually with blocks
-    deformationModelApparentYearScaling: 1,
+    deformationModelApparentYearScaling: .001,
 
     deformSpeedPlate1: 0,     // mm/yr
     deformDirPlate1: 0,       // º from N


sfentress (approved):
This makes it such that in the first deformation simulation (assuming it's still the first) where the user simply runs the model at a given speed, the model will default to running only "500" years.

Screen Shot 2021-09-08 at 1 47 34 PM

I guess no matter what an author will either need to modify the timescale of the first model or the second model, but it feels a little odd to me to set a default timescale that's not the real one, because they will still need to change the distance scale in any case, right? (There's no model where it makes sense to see curvature after 500 years at a scale of 20km).

However, even if this doesn't make sense to me, I'm sure it makes more sense to the team, and in any case the code is fine :+1:

mklewandowski:
> This makes it such that in the first deformation simulation (assuming it's still the first) where the user simply runs the model at a given speed, the model will default to running only "500" years.
> 
> Screen Shot 2021-09-08 at 1 47 34 PM
> 
> I guess no matter what an author will either need to modify the timescale of the first model or the second model, but it feels a little odd to me to set a default timescale that's not the real one, because they will still need to change the distance scale in any case, right? (There's no model where it makes sense to see curvature after 500 years at a scale of 20km).
> 
> However, even if this doesn't make sense to me, I'm sure it makes more sense to the team, and in any case the code is fine 👍

Good questions.  I also thought this seemed a little odd so I ran this by Chris to confirm that we want this as the default in all cases for the Seismic unit.  He confirmed that they did, but he might not have been aware of your comments here.  I will check with him one more time.

PR #265, opened 2021-09-21
Fault orientation

sfentress:
@mklewandowski I realized I somehow never made this into a PR.

This implements fault rotation as a rotation of the entire canvas. This way the grid lines, stations and labels don't all need to have recalculated positions.

There is a little bit of fussy calculations when it comes to placing things like the plate labels, so that they don't rotate out of the square. I tried to keep it to a minimum. You can see how everything moves by opening the authoring and playing with the rotation slider.

Text position and orientation took a while of scratching my head to figure out. Basically, the entire canvas is rotation by xº around the center. Then, when placing the text, we rotate the entire canvas by -xº around the text's location, and then rotate it back.


mklewandowski (approved):
👍 I think this looks good!

mklewandowski:
@sfentress I'm going to merge these changes in hopes of getting the GeoCoder 2.4 build done.

PR #268, opened 2021-10-29
Reload button

sfentress:
This adds a reload button to the button bar, which pops up a modal dialog allowing the user to reset to the initial state.

There were a number of ways to force the reload of the Blockly code, the slightly hacky one I settled on ended up actually being the least hacky and brittle.

The one missing feature is the fact that the reload button is not disabled when the state is initially clean. I tied myself into knots trying to get this to work reliably, but with the way that Blockly keeps asynchronously updating its own XML code during startup, particularly with the recent-ish work to move all the blocks into the viewport (which directly updates the xml), it was virtually impossible to determine when a change had been made by Blockly or the user. I had a large collection of hacks put together, which involved various different comparisons of the state and selectively ignoring parts of those with regexes, before I realized it was an unmaintainable brittle mess. I think that feature is not so important.


src/components/app.tsx
```diff
@@ -593,6 +672,42 @@ export class AppComponent extends BaseComponent<IProps, IState> {
...
+                  Reload Model
+                  <div onClick={hideReloadModal}>x</div>
+                </ModalHeader>
+                <ModalContent>
+                  <div>
+                    Are you sure you want to reload the model?
+                  </div>
+                  <div>
+                    Reloading the model will remove all your work and return the
+                    model to it's original settings.
```
mklewandowski, line 689: "its" original content

sfentress: :+1:

mklewandowski (approved):
👍 Looks good!  I don't think it's worth worrying about the initial state.  This does the job well.

PR #270, opened 2021-11-03
Update max deformation block text

mklewandowski:
This PR updates a Deformation block text from "Compute max deformation with [ ] friction" to "Max deformation calculated based on [ ] friction".

PT Story:
https://www.pivotaltracker.com/n/projects/2231509/stories/179608332

Screen Shot 2021-11-03 at 8 55 20 AM


kswenson (approved):
👍 LGTM -- I think I prefer the old wording, but 🤷 .

PR #281, opened 2022-01-12
180812513 remove block summary from report item

knowuh:
While I was removing things, I cleaned up a little, and added student progress info to the answer-summary.


dougmartin (approved):
Looks good :+1:

PR #284, opened 2022-01-21
180239699 & 180239682: Rename Strain rate, and simplify block arguments.

knowuh:
## Changes:
- Rename compute and display "Strain rate" blocks to compute and display "Deformation build-up"
- Compute deformation has only one argument now (an array of GPS stations)
- Make sure Delaney Triangles and strain rate can be calculated with few GPS stations
- Check for number of GPS stations.
- Add error message: "You must provide at least three GPS stations to compute deformation build-up."
- Remove Map painting method arguments for "Color the map by deformation build-up"

## Stories: 
[#180239682] - Edit "Compute Strain Rate" block
[#180667924] - Make sure Delaney Triangles and strain rate can be calculated with few GPS stations
https://www.pivotaltracker.com/story/show/180667924
https://www.pivotaltracker.com/story/show/180239682

Before:
Screen Shot 2022-01-21 at 4 23 59 PM

After:

Screen Shot 2022-01-21 at 4 25 50 PM
Screen Shot 2022-01-21 at 4 25 17 PM
Screen Shot 2022-01-21 at 4 24 19 PM
Screen Shot 2022-01-21 at 4 26 54 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4813,9 +4813,9 @@
       },
       "dependencies": {
         "follow-redirects": {
-          "version": "1.14.6",
-          "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.14.6.tgz",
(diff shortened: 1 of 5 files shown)


mklewandowski (approved):
👍 LGTM!

PR #285, opened 2022-01-24
180411544 changes to increase deformation block

knowuh:
### Deformation Block:
- The title of this deformation block was changed to "Increase Deformation by deformation build up calculated based on..."
- Units of "(mm/yr)" was added to both input labels.
- Under Plate 1 Speed a line was added that says "direction (degrees) = 0". This shows users that the direction of the plate is locked in this direction.
-  Under Plate 2 Speed a line was added that says "direction (degrees) = 180". This shows users that the direction of the plate is locked in this direction.

Screen Shot 2022-01-24 at 5 14 00 PM

### Deformation key:
- The button that used to say "Show Strain' now reads "Show Deformation"
- The top of the deformation color key now reads to "Deformation build-up (s^-1)" with the -1 being superscript.
Screen Shot 2022-01-24 at 5 12 07 PM
Screen Shot 2022-01-24 at 5 11 53 PM

### Show Deformation build-up value block
- The "show strain rate value" block has been renamed to "Show deformation build-up value".
Screen Shot 2022-01-24 at 5 12 51 PM

**Note:** I am not sure why package-lock.json changed this way. I am not worried about it, but it *is* curious.
**Note:** I will be renaming `Strain*` components next. Because moving and renaming files and imports often results in strange diffs, so I am bundling a few of them following the rewording.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12540,6 +12540,16 @@
             "signal-exit": "^3.0.2"
           }
         },
+        "string-width": {
+          "version": "2.1.1",
+          "resolved": "https://registry.npmjs.org/string-width/-/string-width-2.1.1.tgz",
+          "integrity": "sha512-nOqH59deCq9SRHlxq1Aw85Jnt4w6KvLKqWVik6oA9ZklXLNIOlqg4F2yrT1MVaTjAqvVwdfeZ7w7aCvJD7ugkw==",
+          "dev": true,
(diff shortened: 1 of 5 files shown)


knowuh:
@mklewandowski -- a bunch of small text changes for your review when you have a second.

mklewandowski (approved):
👍 LGTM!

PR #287, opened 2022-01-28
181021708: Rename *Strain-* components to *Deformation-*

knowuh:
We have changed the text labels of the block to use `Deformation build-up` instead of `Strain rate`, but we did not rename the variables or components behind the scenes.

Changes in this PR will make it easier for developers to find the correct blocks in the source code. All changes were made mechanically using a few search and replace string operations across all source files.

These changes are being done in their own PR because it is difficult to track file renaming in diffs. It will make code review easier to do them after the other changes are accepted.

[#181021708]
https://www.pivotaltracker.com/story/show/181021708

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## Version 2.6.0 - release February 16, 2022
+### Features/Improvements:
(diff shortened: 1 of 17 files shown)


mklewandowski (approved):
👍 LGTM!

PR #294, opened 2022-03-09
181497650 update dependencies

lublagg:


package.json
```diff
@@ -111,68 +111,68 @@
...
-    "react-tabs": "^3.0.0",
-    "recharts": "^1.5.0",
+    "react-tabs": "^3.2.3",
+    "recharts": "^1.8.5",
     "screenfull": "^4.2.0",
-    "styled-components": "^5.1.1",
+    "styled-components": "^5.3.3",
     "transformation-matrix": "^1.15.3",
-    "uuid": "^3.3.2",
+    "uuid": "^3.4.0",
```
kswenson, line 175: It'd be worth trying to get to a newer version of `uuid` but check [Upgrading From uuid@3.x](https://www.npmjs.com/package/uuid#upgrading-from-uuid3x) for details.

package.json
```diff
@@ -111,68 +111,68 @@
...
     "script-loader": "^0.7.2",
     "style-loader": "^0.23.1",
-    "ts-jest": "^26.2.0",
+    "ts-jest": "^26.5.6",
     "ts-loader": "^5.3.1",
     "ts-node": "^7.0.1",
-    "tslint": "^5.12.0",
+    "tslint": "^5.20.1",
     "tslint-loader": "^3.6.0",
     "tslint-react": "^3.6.0",
```
kswenson, line 122: Given that `tslint` has been deprecated for some time now, you'll want to move off of these dependencies but you may want to wait for a separate PR for that as it requires additional configuration changes.

package.json
```diff
@@ -111,68 +111,68 @@
     "npm-run-all": "^4.1.5",
     "postcss-loader": "^3.0.0",
     "rimraf": "^2.6.2",
-    "sass-loader": "^7.1.0",
+    "sass-loader": "^7.3.1",
```
kswenson, line 114: This may be a good time to consider switching from `node-sass` to `sass` as well, which is usually a drop-in replacement and doesn't have the compatibility issues that `node-sass` presents due to its binary implementation (`sass` is JavaScript).

PR #298, opened 2022-03-12
Fixing some cypress tests that fail with the latest version of cypress

tejal-shah:


knowuh (approved):
Looks good to me, thankyou

PR #303, opened 2022-03-31
181576799 block inputs

lublagg:


knowuh (approved):
The code looks good to me, thanks.

I thought I found a function that wasn't being used anywhere. If that is true, it would be good to remove it.

PR #304, opened 2022-04-01
181658884 plot deformation block

lublagg:


knowuh (approved):
Looks good, thank you

PR #307, opened 2022-04-04
Earthquake shows on graph as vertical drop

lublagg:


src/components/deformation/block-inputs/block-inputs-menu.tsx
```diff
@@ -64,10 +64,15 @@ export default class BlockInputsMenu extends BaseComponent<IProps, IState> {
...
           color={blockInputButtonColor}
           onClick={this.toggleBlockInputs}
-          disabled={running ? true : false}
+          disabled={running || !deformationHistory.length ? true : false}
           running={running}
         >
-          <ButtonText running={running}>Block Inputs</ButtonText>
+          <ButtonText
+            running={running}
+            disabled={running || !deformationHistory.length ? true : false}
```
knowuh, line 72: I think these logic expressions are unnecessarily complicated.   These are the same I think:

```javascript

> x = false || !0 ? true : false
true
> x = false || !0
true
> x = false || !1 ? true : false
false
> x = false || !1
false

```

src/components/charts/svg-d3-line-chart.tsx
```diff
@@ -107,10 +108,13 @@ export const SvgD3LineChart = (props: LineChartProps) => {
         .text((d) => "Run " + d.group)
         .style("fill", (d) => myColor("run" + d.group) as string)
         .style("text-decoration", "underline")
+        .style("cursor", () => running ? "default" : "pointer")
         .style("font-size", 15)
       .on("click", (d) => {
+        if (!running){
```
knowuh, line 114: minor nitpick here: It might be more legible to me if the body of the if statement was indented.

src/stores/seismic-simulation-store.ts
```diff
@@ -383,6 +390,13 @@ export const SeismicSimulationStore = types
     },
     setPlotOnGraph(){
       const lastGroup = self.deformationHistory[self.deformationHistory.length - 1];
+
+      const firstValueOfLastGroup = lastGroup.values[0];
+
+      if (firstValueOfLastGroup.plotOnGraph === false){
```
knowuh, line 396: A small comment here about intention could be helpful for someone reading without much context.

knowuh (approved):
Looks good, thank you.

Some small suggestions to think about out inline.

PR #308, opened 2022-04-06
Updated axis labels [#181658720]

lublagg:


knowuh (approved):
Looks good to me.

I was just wondering if we should keep the pattern of externalizing all the string sources?

PR #313, opened 2022-04-12
QA fixes

lublagg:
Note: currently, the run button resets several stores when it is hit. Because of this, if a student is stepping through the Create Graph + Run Simulation + Plot Data code blocks, and then presses the run button, it will cause an error, because all of the data that was set initially will be reset. I noticed this same issue happens with the Tephra unit with the Monte Carlo - 3 Locations code.


knowuh (approved):
This looks good to me.  

Though, there is one location where I think the code needs to be re-indented.

👍

PR #333, opened 2022-10-19
[#183583725] Created line break in "max deformation" block.

lublagg:


src/blockly-blocks/deformation/block-deformation-create-sim.js
```diff
@@ -139,6 +139,8 @@ Blockly.Blocks['deformation-create-sim'] = {
     init: function() {
       this.appendDummyInput()
           .appendField(strings.MAX_DEFORMATION)
+      this.appendDummyInput()
```
dougmartin, line 142: I think you can remove this line if you want as it will chain with the previous line.

dougmartin (approved):
Looks good :+1:

PR #341, opened 2024-05-23
Make highlighted deformation line visible no matter what size the graph is (PT-187106210)

lublagg:
Fixes this issue: https://www.pivotaltracker.com/story/show/187106210

This PR changes the deformation graph model so that the amount of overflow from the canvas is calculated dynamically, and so that the number of squares inside the model is always 25. This way, the number of horizontal lines is always proportional to the canvas size and to the overflow, and we can easily dynamically calculate the middle line at which to display the highlight.


eireland (approved):
👍 LGTM

PR #342, opened 2024-05-29
Update S3 deployment.

lublagg:


scytacki (changes requested):
The main changes look good. 

The potential problem is that the project uses the webpack copy plugin to copy the assets from `src/assets` to `dest/assets`. This often means there is some code which is loading assets directly via a URL like: ``. Those kinds of references will break when it is released. More info is here: https://github.com/concord-consortium/starter-projects/blob/main/doc/deploy.md#index-tophtml

And this section of that same file describes how to test for these problems locally:
https://github.com/concord-consortium/starter-projects/blob/main/doc/deploy.md#local-testing-for-compatibility-with-index-tophtml 

If the assets are only used from CSS then they'll probably be fine. The problem is if they are used in the html files or (more likely) in the javascript that is generating the html elements.

Perhaps there won't be any problems, you should test it locally using the pattern above to see if things are looking OK.

src/assets/blockly-authoring/index.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as SeismicGPSGraph from "./toolbox/seismic-gps-graph-toolbox.xml";
+import * as SeismicGPSDeformation from "./toolbox/seismic-gps-deformation-toolbox.xml";
+import * as SeismicGPSEarthquakes from "./toolbox/seismic-gps-deformation-earthquakes-toolbox.xml";
+
+import * as Basic from "./code/basic-setup.xml";
+import * as NestedLoops from "./code/nested-loops.xml";
+import * as MonteCarloCode from "./code/monte-carlo.xml";
+import * as MonteCarlo3Locs from "./code/monte-carlo-3-locs.xml";
+import * as WindDataCollection from "./code/wind-data-collection.xml";
+import * as FilteredWindDataCollection from "./code/filter-wind-data-collection.xml";
```
scytacki, line 19: These can probably all be simplified to remove the `* as`.

scytacki (changes requested):
Nice job. 👍  I found a few things that I think can be cleaned up now that the project is using Webpack 5. Also noted that the report html will not work properly unless it is copied by the release.yml. 

Additionally the release.yml seems to be copying the index-top.html to index-demo.html. If a release is supposed to update the app at: https://geocode-app.concord.org/ then it should copy index-top.html to index.html.

However maybe you want to leave it as index-demo.html for now so you can test out the release process without actually changing https://geocode-app.concord.org/

scytacki (approved):
looks good to me. 👍

PR #343, opened 2024-06-11
Update position over time and velocity data files. (PT-187106362)

lublagg:
This PR makes the following changes:
- All GPS stations in the GeoCoder data set are updated with their recent velocity and direction 
* The GPS stations currently with daily position data are updated so that their daily position data extends to the current times
* Stations around Stockton are added to the number of stations with daily position data


emcelroy (approved):
Looks good to me 👍

PR #344, opened 2024-06-11
Fix: lines were not bending. (PT-187106210)

lublagg:
This PR fixes the bug that was demo'd during Teal Sprint Review today. I had accidentally set one of the arguments to `generateHorizontalLines` to an incorrect value. Now the behavior of the deformation graph is the same as what's on the master branch.


emcelroy (approved):
Looks good 👍

PR #348, opened 2025-04-23
[GEOCODE-24] update more devDependencies

kswenson:
- update cypress to v11 (v12 generated errors related to `isolatedModules`, etc. that I decided not to get hung up on)
- update remaining `devDependencies`


kswenson:
Merging without review as 1. it's just `devDependencies`, 2. @tealefristoe is on vacation, and 3. no one else cares.

PR #351, opened 2025-04-30
chore: update `dependencies-notes.md` with blockly version

kswenson:
Rather than being installed via npm like most libraries, the [blockly](https://github.com/google/blockly) library is simply included in the `geocode` project as minified files in `src/public`: `blockly_compressed.js`, `blocks_compressed.js`, `en.js`, and `javascript_compressed.js`. There is no version number associated with these files, so it was not obvious what version of the `blockly` library was in use. By cross-indexing the date the files were added to `geocode`, the release notes for `blockly` releases around that time, and looking for API changes documented in the release notes in the code itself, I was able to determine that we are using the [April 2019](https://github.com/google/blockly/releases/tag/1.20190419.0) release of the `blockly` library. This PR and the information added to the `dependencies-notes.md` document by this PR document this for use down the road when we want or need to update the `blockly` library.

### DIFF
--- a/dependencies-notes.md
+++ b/dependencies-notes.md
@@ -15,16 +15,16 @@ Notes on dependencies, particularly reasons for not updating to their latest ver
 
 ## Runtime Dependencies
 
-|Dependency          |Current |Latest  |Notes                                                                                   |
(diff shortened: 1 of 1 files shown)


tealefristoe (approved):
👍

PR #353, opened 2025-05-02
Geocode-29 Javascript Molasses Simulation

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/GEOCODE-29

This PR makes a bunch of major updates, mostly to bring the MOLASSES c algorithm (https://github.com/geoscience-community-codes/MOLASSES) to javascript.
- A number of changes have been made to support workers, which are used to load elevation data and run the simulation.
- Elevation data is loaded from an `.asc` file when the unit is `LavaCoder`.
  - Three `.asc` files have been added to `public/data`.
  - `MaunaLoa_30m_UTM5N.asc` is the original file with 30m resolution.
  - `MaunaLoa_60m_UTM5N.asc` and `MaunaLoa_90m_UTM5N.asc` are copies of the same map with lower resolution. We're currently using `MaunaLoa_60m_UTM5N.asc`.
  - `src/scripts/reduce-dem.py` contains a python script that reduces the granularity of a `.asc` file.
  - The elevation data is loaded in `src/components/lava-coder/raster.worker.ts`, a worker that runs separately from the main thread.
- A hack has been added to run the MOLASSES simulation when the user pushes the play button. This should be removed once we've added a code block that will properly run the MOLASSES simulation in the normal way.
- Main MOLASSES lava simulation:
  - The bulk of the work happens in `molasses.ts`, which replicates the c MOLASSES algorithm.
    - The algorithm here mostly stays faithful to the c implementation. One place where it diverges is in how parents are determined to prevent lava from sloshing back to where it came from.
    - This algorithm is run inside a worker and periodically (currently, 5 times a second) posts messages back to the main thread with the status of the lava.
    - A console log is written when the algorithm ends saying how long it took.
  - The simple `molasses.worker.ts` just calls the function to run the simulation.
  - The status of lava is displayed in the 3d map.
    - A mobx `autorun` has been added to `use-cesium-viewer.ts` to add an image layer displaying the current status of the lava to cesium whenever a frame is sent from the algorithm.
      - To prevent flickering, two frames are displayed at any given time. A frame is only removed when it is the third oldest.

Other changes:
- Displaying/hiding labels now only changes the base image layer to avoid interfering with the lava image layers.


src/components/lava-coder/lava-coder-view.tsx
```diff
@@ -33,13 +36,42 @@ export function LavaCoderView({ width, height, margin }: IProps) {
...
+        console.error("Error handling worker message:", error, e);
+      }
+    };
+
+    rasterWorker.postMessage({ type: "start" });
+
+    return () => {
+      rasterWorker.terminate();
+    };
+  }, []);
```
kswenson, line 73: I would move this into its own custom hook, something like `useElevationData()` and then only the return value is needed here. In an upcoming PR I've done something similar for `useHazardZones()` and `useTerrainProvider()`.

src/scripts/reduce-dem.py
kswenson, line 1: Move the scripts folder up so it's a sibling of `src` so that it doesn't get picked up by `tsconfig`, for instance.

src/stores/lava-simulation-store.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Vent near the north, quickly hitting a wall and splitting
+    // ventEasting: 237214,
+    // ventNorthing: 2173722,
+    // Vent near the north east corner of the current map, quickly falling into the ocean
+    // ventEasting: 279000,
+    // ventNorthing: 2181000,
+    totalVolume: 200000000,
+    pulseVolume: 100000, // Standard for small eruption
+    pulseCount: 0,
+    lavaElevations: types.array(types.array(types.number)),
```
kswenson, line 41: In a comment above, you say that we don't store lava elevations in the model, but here we're storing lava elevations in the model. (?) If we are going to store them in the model, consider making them frozen so MST doesn't make every element of the matrix independently observable.

src/stores/lava-simulation-store.ts
```diff
@@ -0,0 +1,98 @@
...
+    raster: types.maybeNull(types.frozen()), // AsciiRaster
```
kswenson, line 42: Isn't this just the contents of the DEM file? Seems like that shouldn't need to be saved in the MST model, either. Perhaps it could be volatile and a description of the DEM file used (e.g. its name) could be stored with the model?

webpack.config.js
```diff
@@ -36,12 +36,26 @@ module.exports = (env, argv) => {
...
+            }
+          ]
+        },
         {
           test: /\.tsx?$/,
           loader: 'ts-loader'
         },
         {
-          test: /\.css$/,
+          test: /\.(css|scss)$/,
```
kswenson, line 58: I don't think adding this extension here is going to work without also adding `sass` and `sass-loader` as dependencies and configuring them in this file, which we may want to do at some point, but I don't think it makes sense to just add the extension.

src/stores/lava-simulation-store.ts
tealefristoe: My thought here was that we would want to include the lava elevations in the model for the purposes of saving/loading the simulation. But given that we're not there yet, and we may not even care about saving the state of the lava anyway, I'm just going to remove this.

webpack.config.js
tealefristoe: I added `sass` and `sass-loader` to the project, and converted `lava-coder-view.css` to `.scss`. Hopefully I did everything correctly, but please double check because I'm not super comfortable in this space.

package.json
```diff
@@ -181,6 +182,8 @@
     "react-syntax-highlighter": "^15.6.1",
     "react-tabs": "^4.3.0",
     "recharts": "^2.15.3",
+    "sass": "^1.88.0",
+    "sass-loader": "^16.0.5",
```
kswenson, line 186: These should be in the `devDependencies` section above.

webpack.config.js
kswenson: LGTM other than having the sass dependencies in `devDependencies` rather than `dependencies` in `package.json`.

src/stores/lava-simulation-store.ts
kswenson: I think it makes sense to leave the model light for now and just add the heavy items if/when we determine a need for them.

kswenson (approved):
👍 Looks good -- just one suggestions to move the sass dependencies to `devDependencies`.

PR #361, opened 2025-05-14
Geocode-8 Lava Blocks

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/GEOCODE-8

This PR adds code blocks that control the MOLASSES simulation.
- Four code blocks have been added: one that includes all parameters, and three that include just a single parameter each (total eruption volume, lava front height, and latitude and longitude of vent).
- These code blocks have been added to the full toolbox but no other toolboxes.
- Fields are prepopulated with defaults but have no range controls set up yet.
- Latitude and longitude are currently separate fields, but this may change in the future.

Additional changes:
- It's no longer possible to specify the vent location using northing and easting. The defaults for these have been removed from the code.
- The play button no longer automatically starts the simulation when in the LavaCoder unit. It now works as normal in all situations, relying on a MOLASSES code block to run the simulation.


kswenson (approved):
👍 LGTM -- just a couple things to consider.

PR #362, opened 2025-05-15
Geocode-16 Lava Widgets

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-16

This PR adds two widgets to the Molasses simulation mode of GeoCode, Lava Front Height and Eruption Volume. Note that Eruption Volume isn't technically part of the story, but it doesn't look like there is another story for it--the project team apparently assumed that it would be possible to reuse the existing widget for this purpose with no extra work required, which was not the case. In particular:
- The scaling is different between Molasses and Tephra, the other simulation that uses the widget.
- Colors are different between the two simulations.

Unrelated, a small change was made to convert the elevation data from feet to meters. Previously, the base elevation data was in feet, but the lava elevation is in meters. Now they use the same unit, making the simulation more accurate. This change does impact the simulation, making the resulting lava flows spread much more.


cypress/e2e/smoke/conditions-tab-ui.cy.js
```diff
@@ -31,5 +33,11 @@ context("Conditions panel", () => {
             map.getKeyContainer().should('contain','Risk Level');
             map.getMapKeyToggle().should('contain','Show Tephra');
         });
+        it('verity LavaCoder shows correct elements',()=>{
+            modelOptions.getModelOptionsMenu().click();
+            modelOptions.selectUnitOption('LavaCoder');
+            conditionsTab.getLavaFrontHeightWidget().should('be.visible');
+            conditionsTab.getVolumeOfLavaWidget().should('be.visible');
+        });
```
kswenson, line 41: This is fine, but I think there's also a way to specify the simulation via url parameter which would obviate the need to navigate the model options menu.

src/components/lava-coder/lava-constants.ts
```diff
@@ -6,6 +9,13 @@ export const defaultResidual = 5;
 export const defaultVentLatitude = 19.5;
 export const defaultVentLongitude = -155.565;
 
+export const minEruptionVolume = 1000000;
+export const maxEruptionVolume = 10000000000;
+
+export const minResidual = 2;
+export const maxResidual = 50;
```
kswenson, line 16: Really? 50 seems excessive, but what do I know. 🤷‍♂️

src/components/lava-coder/parse-ascii-raster.ts
```diff
@@ -23,7 +24,7 @@ export function parseAsciiRaster(content: string) {
 
   // Parse raster values
   const values = lines.slice(dataStartIndex).map(line =>
-    line.trim().split(/\s+/).map(Number)
+    line.trim().split(/\s+/).map(value => Number(value) * kMetersPerFoot)
```
kswenson, line 27: Perhaps a comment about the DEM values being in feet.

src/components/widgets/lava-front-height-widget.tsx
```diff
@@ -0,0 +1,46 @@
...
+import LavaFront1 from "../../assets/lava-coder/widget-lava-front-height-1.png";
+import LavaFront2 from "../../assets/lava-coder/widget-lava-front-height-2.png";
+import LavaFront3 from "../../assets/lava-coder/widget-lava-front-height-3.png";
+import LavaFront4 from "../../assets/lava-coder/widget-lava-front-height-4.png";
+import LavaFront5 from "../../assets/lava-coder/widget-lava-front-height-5.png";
+import LavaFront6 from "../../assets/lava-coder/widget-lava-front-height-6.png";
+import LavaFront7 from "../../assets/lava-coder/widget-lava-front-height-7.png";
+import LavaFront8 from "../../assets/lava-coder/widget-lava-front-height-8.png";
+import LavaFront9 from "../../assets/lava-coder/widget-lava-front-height-9.png";
+import LavaFront10 from "../../assets/lava-coder/widget-lava-front-height-10.png";
```
kswenson, line 15: We never display `widget-lava-front-height-none.png`?

src/components/widgets/lava-front-height-widget.tsx
```diff
@@ -0,0 +1,46 @@
...
+  lavaFrontHeight: number;
+}
+export function LavaFrontHeightWidget({ lavaFrontHeight }: LavaFrontHeightWidgetProps) {
+  const iconIndex = Math.floor((lavaFrontHeight - minResidual) / rangeResidual * lavaFrontIcons.length);
+  const lavaFrontIcon = lavaFrontIcons[Math.max(0, Math.min(iconIndex, lavaFrontIcons.length - 1))];
+
+  return (
+    <ValueContainer backgroundColor={kWidgetPanelInfo.right.backgroundColor}>
+      <RelativeIconContainer>
+        <img src={lavaFrontIcon} />
```
kswenson, line 39: Perhaps some alt text, e.g. `Lava front height: x`?

kswenson (approved):
👍 LGTM -- just a few questions/suggestions.

cypress/e2e/smoke/conditions-tab-ui.cy.js
tealefristoe: I'm going to put a TODO in about this but I'm not going to change it for now.

src/components/lava-coder/lava-constants.ts
tealefristoe: Agreed, but that's what's specified in the Jira stories. I suspect that once the researchers get a chance to play around a bit they'll want to make a few changes.

PR #365, opened 2025-05-18
Fix Loading DEMs

tealefristoe:
This PR fixes an issue where elevation data could not be loaded on deployed branches.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -122,7 +122,7 @@ module.exports = (env, argv) => {
           type: 'asset/source'  // Exports the raw source code of the file
         },
         {
-          test: /\.(kml|kmz|xml)$/i,
+          test: /\.(asc|kml|kmz|xml)$/i,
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -62,7 +62,7 @@ module.exports = (env, argv) => {
           ]
         },
         {
-          test: /\.(png|woff|woff2|eot|ttf)$/,
+          test: /\.(png|woff|woff2|asc|eot|ttf)$/,
```
kswenson, line 65: You should only need to list `asc` once in this file, and I think it should be the `asset/resource` section below, so you should be able to remove it from here.

kswenson (changes requested):
👍 Looks good -- just remove `asc` from the `asset` section (and verify that everything still works).

kswenson (approved):
👍 LGTM

PR #366, opened 2025-05-18
feat: vent location dialog; vent location marker handles vertical exaggeration [GEOCODE-18]

kswenson:
[[GEOCODE-18](https://concord-consortium.atlassian.net/browse/GEOCODE-18)] Add "Place Vent" button

- show vent location dialog with button for copying coordinates to clipboard
- vent location dialog is dismissed when clicking anywhere outside of it
- clicking on vent location marker shows vent location dialog again
- separately, vent location marker and vent location dialog handle vertical exaggeration properly

[GEOCODE-18]: https://concord-consortium.atlassian.net/browse/GEOCODE-18?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/lava-coder/vent-location-popup.tsx
```diff
@@ -0,0 +1,98 @@
...
+import CopyButtonIcon from "../../assets/lava-coder/content-copy-icon.png";
+import IconButton from "../buttons/icon-button";
+
+import "./vent-location-popup.scss";
+
+interface IProps {
+  viewer: CesiumWidget | null;
+  ventLocation: Cartographic | null;
+  verticalExaggeration: number;
+  show: boolean;
```
tealefristoe, line 13: `show` seems a bit confusing to me. `visible`? `hidden`?

tealefristoe (approved):
Looks great 👍 

I left one minor suggestion to consider. I also wonder if we should include all the various icon files in the repo, even though we're only using 1/4 of them?

src/components/lava-coder/vent-location-popup.tsx
kswenson: It's fairly standard for popups and dialogs, so I think I'll leave it for now.

kswenson:
> I also wonder if we should include all the various icon files in the repo, even though we're only using 1/4 of them?

I'm somewhat conflicted about this as well. The reason to use one of the other .png or .svg files is to perform better on hi-DPI devices. They don't work as drop-in replacements, however, and I didn't think it was worth it to take time to deal with it now, so I'm leaving them in the repo as a reminder to revisit the situation when time permits (which may well be never). But it seems like having them in the repo makes it at least slightly more likely that it will be addressed at some point than just having them in the spec document so I'm inclined to leave them for now.

PR #372, opened 2025-05-22
Geocode-41 Full Map

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-41

This PR uses a new DEM map that covers most of the large island.
- The old, smaller map can still be used by setting `useLargeMap` in `lava-options.ts` to `false`.
- If `renderMapExtent` in `lava-options.ts` is set to `true`, the area covered by the large map will be rendered on the 3d model when the simulation is run. This was useful for positioning the large map, and it seemed like functionality that might come in handy again at some point in the future.
- The project team has requested that we use feet for the elevation data rather than meters. This is technically incorrect, but the team likes the results better. So in this PR, I'm no longer converting the old smaller map from feet to meters, and I'm even converting the new larger map from meters to feet.


src/components/lava-coder/lava-options.ts
```diff
@@ -0,0 +1,8 @@
+// This flag controls which elevation map is used for the simulation.
+// When true, the larger map that covers almost all of the large island is used.
+// When false, the smaller map that only covers the central eastern part of the island is used.
+export const useLargeMap = true;
```
kswenson, line 4: Seems like this should be authorable, but that needn't hold up this PR.

src/components/lava-coder/lava-options.ts
```diff
@@ -0,0 +1,8 @@
...
+
+// Set this to true to render the full elevation map extent when the eruption starts.
+// This does not impact the simulation, but it's helpful for visualizing what is and is not included in the map.
+export const renderMapExtent = false;
```
kswenson, line 8: If this were authorable, then it could be useful to the project team.

kswenson (approved):
👍 Looks good -- I don't like the subterfuge with meters/feet of elevation, but that's not our call. I pointed out a couple of places where it might make sense to make parameters authorable rather than hard-coding them, but that need not hold up the PR.

PR #373, opened 2025-05-22
Geocode-43 Clear Old Simulation

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-43

This PR clears the lava from an old simulation when a new simulation starts.
It also prevents a new simulation from starting when an old simulation is currently running.


tealefristoe:
@kswenson I just requested your review on this PR, but looking more closely at the Jira story it needs more work. I'll ping you when it's ready.

tealefristoe:
@kswenson Alright this is ready now.

src/stores/lava-simulation-store.ts
```diff
@@ -99,10 +101,6 @@ export const LavaSimulationStore = types
         ventLongitude: self.ventLongitude
       };
       self.worker.postMessage({ type: "start", parameters });
-
-      return () => {
-        self.worker?.terminate();
-      };
```
kswenson, line 105: Seems like we would still want the worker to be terminated when the component is unmounted, so I would think we still want to return this termination function from the `useEffect`.

kswenson (approved):
👍 Looks good -- just one suggestion to restore the disposer function for the useEffect.

src/stores/lava-simulation-store.ts
tealefristoe: This is actually not in a `useEffect`, I think it was at some point during development, which is why this return statement was added. But it hasn't been doing anything for a while now.

PR #382, opened 2025-05-29
Geocode-19 Vent Key

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-19

This PR adds a vent key button. The button is only visible when the hazard map is displayed.


kswenson (approved):
👍  LGTM

PR #388, opened 2025-06-02
feat: molasses performance tweaks

kswenson:


tealefristoe (approved):
Looks good 👍 

Did you try only creating the neighbors when a cell becomes active, rather than creating the neighbors for all cells at the beginning? Did that make a meaningful difference?

PR #391, opened 2025-06-02
feat: `Reload` button respects `unit` url parameter; vent location marker enhancements

kswenson:
[[GEOCODE-51](https://concord-consortium.atlassian.net/browse/GEOCODE-51)] Reload button on LavaCoder should not reload to Tephra unit

- feat: reload button respects `unit` url parameter
- feat: hide vent location marker during simulation
- feat: Place Vent button disabled during simulation
- feat: vent location marker resets when reset button clicked
- feat: show initial/default vent location marker

While working on the `Reload` button behavior, I noticed that the vent location marker wasn't being reset correctly on reset/reload. This was because the vent location was being stored redundantly in the model and in state and only the model was being reset. While working on the vent location marker, requests came in from the project team to hide the vent location marker during the running of the simulation and to disable the `Place Vent` button during the simulation. And while working on those I noted a few other things along the way.

[GEOCODE-51]: https://concord-consortium.atlassian.net/browse/GEOCODE-51?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/lava-coder/lava-coder-view.tsx
+++ b/src/components/lava-coder/lava-coder-view.tsx
@@ -1,6 +1,6 @@
 import { reaction } from "mobx";
 import { observer } from "mobx-react";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import VentLocationMarkerIcon from "../../assets/lava-coder/location-marker.png";
 import { lavaSimulation } from "../../stores/lava-simulation-store";
 import { LavaMapType, LavaMapTypes, uiStore } from "../../stores/ui-store";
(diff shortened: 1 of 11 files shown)


src/components/lava-coder/lava-coder-view.tsx
```diff
@@ -104,12 +119,12 @@ export const LavaCoderView = observer(function LavaCoderView({ width, height, ma
                 "elevation:", `${Math.round(elevation)}m = ${elevationFeet}ft`,
                 "in hazard zone:", isInHazardZone);
     if (isPlaceVentMode && isInHazardZone) {
-      setVentLocation(latitude, longitude, elevation / verticalExaggeration);
+      lavaSimulation.setVentLocation(latitude, longitude, elevation / verticalExaggeration);
```
tealefristoe, line 122: With this change, the simulation automatically updates its vent location on vent placement. But the blockly block won't update its vent location value automatically, so unless the user manually pastes the new location in the block, that value will override the pinned location when the simulation is run. I guess that's how things work right now anyway, but I really hope we can figure out a better solution for all this at some point.

tealefristoe (approved):
Looks good 👍

src/components/lava-coder/lava-coder-view.tsx
kswenson: In other words, previously there were _three_ independent representations of the vent location: the model, React state, and the Blockly block. This PR reduces it to two by eliminating the React state, but you're right there's still the Blockly block to consider. But now that I think about it, I think that's actually the right behavior. The Blockly blocks represent the "program" the student wrote. That program sets the model state when it is run. Since the story specifies that reload doesn't affect the Blockly blocks, I think this is the right behavior, even if it's potentially confusing.

PR #394, opened 2025-06-03
feat: remove default vent location marker

kswenson:


tealefristoe (changes requested):
Right now, the pin reappears after a simulation run if the vent is not at the default location. Chris requested that the pin not show up after a run.

I also noticed that when you start a run with the location popup visible, it remains visible for a second or two after the pin is removed. If the pin and popup are at a different location than the vent, the popup flashes to the vent location before it disappears. I think both the pin and popup should be removed at the same time.

src/stores/lava-simulation-store.ts
```diff
@@ -48,9 +49,15 @@ export const LavaSimulationStore = types
...
     get isRunning() {
       return self.worker != null && self.pulseCount < uiStore.pulsesPerEruption;
+    },
+    get isDefaultVentLocation() {
+      return self.ventLatitude === defaultVentLatitude && self.ventLongitude === defaultVentLongitude;
     }
   }))
   .views((self) => ({
+    get showVentLocationMarker() {
+      return self.ventLocationChanged && !self.isDefaultVentLocation;
```
tealefristoe, line 59: Do we still need to check for the default location here? Seems like we can get rid of that check and with it the view.

tealefristoe (approved):
Looks good 👍 

Just one question to consider.

src/stores/lava-simulation-store.ts
kswenson: I think if you set a vent location and then run the model without copying the new location to the block, that blockly will reset the vent location in the model to the default, which will set the ventLocationChanged flag, so I'm inclined to think that the default location test is still necessary.

tealefristoe: This sounds wrong to me. Imagine the user changes the vent location manually so it's no longer default, then places a pin, then starts the simulation. That will change the vent location to the first non-default spot. Will the pin show up then?

If I'm not mistaken, the only time a pin should show up is when the user clicks to drop it. The pin shouldn't show up whenever the location changes, it should show up in the click handler.

kswenson: > This sounds wrong to me. Imagine the user changes the vent location manually so it's no longer default, then places a pin, then starts the simulation. That will change the vent location to the first non-default spot. Will the pin show up then?

No, because the marker is always hidden during (and after) running the simulation.

> If I'm not mistaken, the only time a pin should show up is when the user clicks to drop it. The pin shouldn't show up whenever the location changes, it should show up in the click handler.

I just tested these scenarios and I believe it is working as desired. In the scenario of placing a pin and then running with the default vent location there is a point at which the `isDefaultVentLocation` is what's keeping the marker from showing, perhaps only briefly. I'm going to leave this for now and we can revisit when next we're working on it. One of the paths I went down before settling on the current approach is to store the user-set vent location in the `uiStore` and then only the running of the simulation would set the _actual_ vent location in the model. I think that's actually a better representation of what's going on, but that was starting to have more far-reaching consequences than seemed warranted so I saved it in a branch, but we may want to go back to that approach at some point.

tealefristoe: If the pin doesn't show up in the scenario I described, I don't see why it would show up in the scenario where the default vent is used. I just tried without the default check, and I didn't see the pin pop up. We can discuss and try it in the office if you want.

I'm going to make a PR that addresses this now. If you and I work on it down the road, we might remember what happened here and be able to fix it without having to relearn what's going on. If other engineers work on it, this will be mysterious and they will either spend a bunch of time figuring out what's going on, or will just trust that it's right without understanding it and be afraid to change it.

PR #415, opened 2025-08-18
GEOCODE-75 Arithmetic Block

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-75

This PR adds the built-in arithmetic block to the data category for LavaCoder.

It also fixes the colors for blockly blocks. These broke when we updated blockly to the latest version.


tealefristoe:
@scytacki I'm not sure of a way to print a variable, but you can check the resulting javascript code.
- In the model options in the top right, choose Tephra as the unit.
- In the model options under Left Tabs, check Show code?
- Change the unit to LavaCoder.
- Add the blocks you want to test.
- Choose the Code tab on the left side of the screen.

This won't confirm that the arithmetic actually works, but it should show that the javascript code it generates looks correct. Given that the arithmetic block is built in to blockly, we'd have bigger problems if it didn't work correctly 😬

PR #416, opened 2025-08-18
GEOCODE-58 Synchronous Eruption

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/GEOCODE-58

This PR uses `js-interpreter`'s `createAsyncFunction` to make the blockly code execution pause until a lava simulation has completed.


scytacki (approved):
Looks good to me.

I did notice that if the blocks are not connected together, the eruptions do not run in parallel. I'm guessing that is not a required feature.

PR #423, opened 2025-08-25
chore: remove widgets panel [GEOCODE-85]

kswenson:
[[GEOCODE-85](https://concord-consortium.atlassian.net/browse/GEOCODE-85)] Remove Widgets from "Map" tab

Note that I have _not_ removed the code for rendering the LavaCoder widgets in this PR as I wouldn't be shocked if they come back in some form at some point. The code for rendering them has merely been disabled so it's easy to restore them if the need should arise.

[GEOCODE-85]: https://concord-consortium.atlassian.net/browse/GEOCODE-85?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -340,10 +340,8 @@ export class AppComponent extends BaseComponent<IProps, IState> {
     const scenarioData = (Scenarios as {[key: string]: Scenario})[scenario];
     const simMargin = { x: 5, y: 5 };
     const simMarginStr = `${simMargin.y}px ${simMargin.x}px 0 ${simMargin.y}px`;
-    // the widgets panel has been removed from LavaCoder, so the map expands to fill the space
-    const kWidgetsPanelHeight = 160;
-    const lavaCoderHeight = mapHeight + kWidgetsPanelHeight - 2 * simMargin.y;
     const tabHeight = height - 30; // Excludes tab button height
+    const mapLavaCoderHeight = tabHeight - 2 * simMargin.y;
     const tableHeight = tabHeight * 0.35;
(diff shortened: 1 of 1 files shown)


src/components/app.tsx
```diff
@@ -340,7 +340,9 @@ export class AppComponent extends BaseComponent<IProps, IState> {
     const scenarioData = (Scenarios as {[key: string]: Scenario})[scenario];
     const simMargin = { x: 5, y: 5 };
     const simMarginStr = `${simMargin.y}px ${simMargin.x}px 0 ${simMargin.y}px`;
-    const lavaCoderHeight = mapHeight - 2 * simMargin.y;
+    // the widgets panel has been removed from LavaCoder, so the map expands to fill the space
+    const kWidgetsPanelHeight = 160;
+    const lavaCoderHeight = mapHeight + kWidgetsPanelHeight - 2 * simMargin.y;
```
tealefristoe, line 345: This section is becoming a bit of a mess. I think we should try to clean it up:
- Define constants for widget and tab button (is there a better name for this?) height.
- Define `mapHeight` and `tabHeight` in terms of `height` and the above constants.
- `lavaCoderHeight` should then be `tabHeight - 2 * simMargin.y`.

tealefristoe (approved):
Looks good 👍 I think it's wise to not delete this code yet.

I do think we should try to clean up the code that defines the heights for the right panel.

kswenson:
> I do think we should try to clean up the code that defines the heights for the right panel.

Done

PR #461, opened 2026-08-18
Geocode-156 Bing Alternatives Spike

tealefristoe:
GEOCODE-156

This PR adds a document with the findings of a spike into cheaper alternatives to Bing satellite imagery for LavaCoder.

It also fixes a bug where Cesium terrain was unnecessarily being instantiated multiple times.

### DIFF
--- a/docs/lavacoder/lavacoder-optimizations.md
+++ b/docs/lavacoder/lavacoder-optimizations.md
@@ -182,3 +182,143 @@ This should happen before any of these options is worth planning in detail.
 - [USGSImageryOnly service](https://basemap.nationalmap.gov/arcgis/rest/services/USGSImageryOnly/MapServer)
 - [USGS Imagery Only catalog entry](https://www.sciencebase.gov/catalog/item/544172dae4b0b0a643c73c6e)
 - [OSM Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/)
+
+---
+
+## Terrain Mesh Alternatives (LavaCoder / Cesium)
+
+Status: exploratory. No decision made. Effort figures are estimates, not commitments.
+
+### Before doing anything
+
+**What is Ion terrain actually costing today?** The streaming estimate below is derived, not measured.
+If real usage is well inside quota, this work may be lower priority than the imagery question.
+
+### The problem
+
+Imagery is only half of our Ion dependency. `createWorldTerrainAsync()` in
+`src/hooks/lava-coder/use-terrain-provider.ts:11` pulls Cesium World Terrain — the 3D *shape* of the
+ground, as opposed to the photo draped over it. **Replacing Bing does not remove the Ion token or the
+metering.**
+
+It is a different meter, which cuts both ways. Terrain is a Cesium-owned asset, so it bills against
+**data streaming GB/month** (15 GB on Community, 150 GB on Commercial) rather than Global Imagery
+sessions. That means it is not subject to the hard per-session wall described above — but it is also
+large enough to matter on its own:
+
+| Sessions/month | Est. terrain streaming (at 1–4 MB/session) | Against quota |
+|---|---|---|
+| 5,000 | ~10 GB | nearly exhausts Community's 15 GB |
+| 100,000 | ~200 GB | exceeds Commercial's 150 GB |
+
+So "drop Bing and fall back to the free Community tier" does not eliminate Ion on its own.
+
+### What the terrain provider is actually used for
+
+| Job | Where | Needs a global service? |
+|---|---|---|
+| Visual 3D relief | `src/hooks/lava-coder/use-cesium-viewer.ts:31` | yes, or a replacement mesh |
+| Camera zoom clamp (stay 1 km above ground) | `src/hooks/lava-coder/use-camera-controls.ts:201` | **no** |
+| Flag placement on the surface | `src/hooks/lava-coder/use-flag-locations.ts:37` | **no** |
+
+The last two go through `getElevation()`, which calls `sampleTerrainMostDetailed` — fetching terrain
+tiles at maximum available detail on *every* zoom step and *every* flag drop. That is a meaningful
+share of the streaming bill spent on what are only point lookups.
+
+### Choosing a source DEM
+
+- **`big_island.asc`** — the raster the simulation already runs on — is the obvious candidate, but it is 60 m posts
+  (2151 × 2448 over the AOI). Cesium World Terrain over Hawaiʻi is roughly 10–30 m effective, so building relief from
+  it would be a 2–4× downgrade, and the 3× default vertical exaggeration would amplify that into visible terracing
+  on the shield slopes.
+- **SRTM 1-arc-second (~30 m)** — the four GeoTIFFs already committed under
+  `src/assets/lava-coder/elevation-maps/` (`n18_w156`, `n19_w155`, `n19_w156`, `n20_w156`). Zero new
+  downloads.
+- **USGS 3DEP 1/3-arc-second (~10 m)** — public domain, covers Hawaiʻi, and would be *better* than
+  what Ion gives us today.
+
+### Options under consideration
+
+Cesium will not consume a `.asc` directly; it wants tiles or a heightmap callback. That constraint
+shapes all three options.
+
+#### Move the point lookups to a local DEM
+
+Rewrite `getElevation()` as a bilinear sample into the raster the worker already loads
+(`src/simulations/lava-coder/raster.worker.ts`). Removes two of the three Ion terrain uses, and with
+them all per-zoom and per-flag network traffic. The lookups become synchronous and free.
+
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 Looks great -- thorough investigation and impressive write-up. The following review, developed in conjunction with Claude Code 🤖, identifies some issues worthy of consideration before merging. If it were me, I'd at least fix the radian/degrees bug, but I leave all of it to your discretion.

---

## PR Review Summary

**Changes:** 6 files, +659 / -15 lines (634 of the additions are a single design document)

### What it does

Two independent things. The bulk is `docs/lavacoder/lavacoder-optimizations.md` — a spike write-up
covering three cost questions for LavaCoder: replacing Bing satellite imagery, replacing Cesium World
Terrain, and sourcing place-name labels. Each section states the problem, enumerates options with
measured evidence, and ends with a comparison table and open questions.

The code change is a small refactor: `useTerrainProvider()` was being called from three separate
hooks, so every session ran `createWorldTerrainAsync()` three times and held three independent
terrain providers. It is now called once in `LavaCoderView` and the results injected downward.

### The approach

`useTerrainProvider` is hoisted to `lava-coder-view.tsx:60`, and its two outputs are passed to the
consumers that used to instantiate their own:

- `terrainProvider` → `useCesiumViewer(container, mapType, terrainProvider)` — positional third arg
- `getElevation` → `useCameraControls({...})` and `useFlagLocations({...})` — both via props objects

`useCameraControls` is converted from positional args to a props interface in the process, matching
the convention `useFlagLocations` already used. A `GetElevation` type is exported from
`use-terrain-provider.ts` so consumers can type the injected function without importing the hook, and
the definition site uses `useCallback<GetElevation>` so parameter types are inferred rather than
repeated.

`git grep useTerrainProvider` on `master` confirms three call sites; on this branch there is exactly
one.

### Assessment

The refactor is correct and the right shape — single owner, inject downward. Verified:

- All former `useTerrainProvider` consumers are converted; no call site was missed.
- Dependency arrays are updated correctly. `getElevation` is a `useCallback` keyed on
  `terrainProvider`, so its identity changes when terrain resolves, and `use-flag-locations.ts:63`
  lists it — flags re-place once real elevations are available.
- `zoomIn`/`zoomOut` are plain per-render functions used only as click handlers, so injecting
  `getElevation` introduces no stale-closure risk.
- The `useCesiumViewer` create/destroy effect behaves as before: it no-ops while `terrainProvider` is
  null and builds the widget on the transition, same as when it owned the provider itself.
- `tsc --noEmit` and `eslint` on the five changed source files both pass clean.

Worth saying the fix is more than tidiness: per the doc's own terrain section, Ion terrain bills
against streaming GB/month, so three redundant provider initializations per session were consuming
real quota.

On the document, I checked the numbers rather than taking them on trust, and they hold up:

- **Every cost table recomputes exactly.** Esri tile model ($0 / $450 / $3,450) and session model
  ($16 / $76 / $396); the A2 worked examples ($156 vs $525, and $36 vs free); and the stated 8–15k
  crossover matches 2M tiles ÷ 250 = 8,000 sessions.
- **Every capacity figure reproduces.** The AOI bounds in `lava-constants.ts` give 125.3 × 150.8 km
  against the stated ~126 × 151. `big_island.asc`'s header confirms `ncols 2151 / nrows 2448 /
  cellsize 60`, and the 5.3M / 21M / 189M post counts and 11 / 42 / 378 MB sizes all follow. Pyramid
  tile counts quadruple per zoom (4.00x, 3.99x). The egress and terrain-streaming tables each imply a
  single consistent per-session rate sitting inside their stated ranges, and $0.085/GB reproduces
  both the $3 and $60 figures.
- **Code and asset references are accurate.** `verticalExaggeration` really does default to 3; the
  keyless ArcGIS `World_Topo_Map` call is where the doc says; `@cesium/engine` is 17.0.0 and does
  export `CustomHeightmapTerrainProvider`; all four named SRTM GeoTIFFs are present.

The write-up is unusually well-evidenced for a spike — candidates were probed directly rather than
assessed from documentation (tile weights measured over six tiles, NOAA coverage verified three ways
against control locations, USGS z17 404s confirmed), and it is explicit about which numbers are
measured and which are derived.

### Issues

**1. The doc diagnoses a real bug on a line this PR edits, and the PR leaves it in place.**
`src/hooks/lava-coder/use-camera-controls.ts:206` — medium severity, pre-existing.

The "Note on a related bug" section is correct; I verified it. `zoomIn` passes
`viewer.camera.positionCartographic.longitude/latitude`, which Cesium returns in **radians**, to
`getElevation`, which calls `Cartographic.fromDegrees`. The zoom-in clamp therefore samples elevation
somewhere in the Atlantic near (-2.7, 0.34), gets ~0, and measures the 1 km floor from sea level
instead of from terrain. Zooming in over Mauna Loa can drive the camera far closer to the ground than
intended.

The doc explicitly defers this ("whoever rewrites `getElevation()` should fix it in the same pass").
But this PR already rewrites the function's signature and touches `getElevation`'s plumbing, the
diagnosis is done, and the fix is one line — `CSMath.toDegrees(...)` on both arguments, with `CSMath`
already imported in the file. Your call whether to take it here or leave it for the rewrite.

**2. The TLDR and the body disagree about whether a decision has been made.** Medium.

TLDR (line 11): "We've explored options and **decided** hosting our own tiles (option B) using the
`SoH_Imagery/Vivid_2022` imagery is the best option."

Body, Comparison (line 348): "Two real paths remain, and they are **a genuine trade rather than a
ranking**... Below roughly 8–15k combined sessions/month Esri is free and B is hard to justify."

Body, Open questions (line 359): "**What is our actual session and tile volume?** This is now the top
blocker: **it decides A-vs-B**."

All three section headers also still read "Status: exploratory. No decision made." A reader who skips
the TLDR reaches the opposite conclusion from one who reads it. If B is decided, it would help to
record what made the volume question stop being the blocker. Related and smaller: the open question
"Will the state license `Vivid_2022` / `WV2_2016`? One email to `gis@hawaii.gov`" is stale, since the
TLDR says that email has been sent.

**3. `WV2_2016` carries three different verdicts.** Medium.

- Candidates table (line 101): "**Viable but heavy** — 10.6x Esri's bandwidth"
- Section heading (line 145): "Hawaii Statewide GIS: **viable**, but the tiles are ten times heavier"
- Bolded verdict directly under that heading (line 147): "**Quality probably too low**"

Line 98 also says "All four candidates have now been probed directly. **Only two are viable**," which
holds under the first two readings but not the third. And option B's source table still lists
`WV2_2016` as a viable self-hosting source without the quality caveat — worth saying whether the
concern carries over there, since self-hosting was going to re-encode those tiles anyway.

The annotation convention is uneven in a way that probably caused this: NOAA gets a bolded
**RULED OUT**, Hawaii gets a bolded status line that contradicts its own heading, USGS and Esri get
neither. One status line per candidate would keep this from drifting again.

**4. Minor, take or leave.**

- **Doc line references drifted within this same PR.** The doc cites `use-camera-controls.ts:201` for
  the `getElevation` call and `:20-22` for the eye-height clamp; after this PR's edits to that file
  they are at `:206` and `:21,23`, and `lava-coder-view.tsx:147-158` is now `:151-162`. Anchoring to
  symbol names (`zoomIn()`, `kMinDistanceAboveTerrain`, `toggleMapType()`) would stay valid.
- **"2–4x downgrade" should be "2–6x"** (line 434). 60 m against "roughly 10–30 m effective" is 2x at
  the coarse end and 6x at the fine end. Doesn't change the conclusion.
- **Lowercase "claude"** at line 17 — looks like it just got missed by the "Improve capitalization"
  commit.
- **Style drift:** `Hawaii` vs `Hawaiʻi` is mixed in prose (14 vs 3) while `Kīlauea` is consistently
  macronned, so the okina omission reads as accidental rather than a deliberate ASCII choice.
  British spellings ("colour" x3, "productionise") appear only in the Labels section. The TLDR is
  unwrapped at 146–212 chars while the body wraps at ~100.
- **Option C cites an unmerged *local* `low-res` branch** — it isn't on origin, so nobody else can
  look at it. Either push it or drop the reference.

**5. Question, not a finding.** The doc says option C (gate Bing behind the existing authoring flag)
"remains worth doing immediately regardless — it is hours of work and stops the bleeding." It isn't
in this PR, and nothing here claims it is: production still gets Bing on first paint
(`defaultMapType()` returns `"terrain"` off-localhost, which falls through to
`createWorldImageryAsync({ AERIAL })`), and `showMapTypeTerrain` / `showMapTypeLabeledTerrain` both
still default to `true`. Is that queued as a follow-up, or is it waiting on the A-vs-B decision, or should it just be done here?

tealefristoe:
@kswenson Thank you for the review. The bug was very easy to fix, so I'm glad you pushed on it 👍 I also fixed up the document with most of the other suggestions.

tealefristoe:
@kswenson In regards to your 5. question, I didn't bother doing that change because we'll implement option B to change the imagery and do away with Bing imagery for good before this simulation sees any more significant use.

